PR #3, opened 2022-09-19
Fix inference job default output table name

smurching:
Fix the default output table name for batch inference; only alphanumeric characters are allowed in unescaped Spark table names, so use the alphanumeric version of the project name in the output table name

Before:
image

After:
image


jerrylian-db (approved):
LGTM

PR #5, opened 2022-09-20
Use official MLflow release

smurching:
Use official MLflow release, instead of MLflow from master, in our tests/production jobs now that the bugfixes in MLflow 1.29.0 are released


mohamad-arabi (approved):
LGTM, thanks

PR #6, opened 2022-09-20
fix batch inference

yinxi-db:
add Jinja2==3.0.3 to requirements.txt 
This way, the batch inference job can succeed.


smurching (approved):
LGTM! I manually verified this fix was necessary for batch inference with MLP to work

PR #14, opened 2022-10-17
White background image for dark-mode readability.

vadim:
When using GitHub in Dark Mode, the black text is not visible due to the transparent background. Set white background.


smurching (approved):
LGTM

PR #16, opened 2022-11-01
remove enterprise=true flag

niall-turbitt:
- enterprise=true feature tag for service principals not necessary


smurching (approved):
LGTM

PR #17, opened 2022-11-03
Pinning mlflow to not pick the latest mlflow2.0 release

sunishsheth2009:
Signed-off-by: Sunish Sheth <sunishsheth2009@gmail.com>

### DIFF
--- a/{{cookiecutter.project_name}}/requirements.txt
+++ b/{{cookiecutter.project_name}}/requirements.txt
@@ -1,4 +1,4 @@
-mlflow[pipelines]>=1.29.0
+mlflow[pipelines]==1.29.0
 numpy>=1.23.0
 pandas>=1.4.3
 scikit-learn>=1.1.1


vladimirk-db (approved):

smurching:
@sunishsheth2009 thanks for the PR, looks good! Do we plan to update the example ML code in the stacks to use MLflow 2.0 once it's out?

mengxr:
@sunishsheth2009 https://github.com/databricks/mlops-stack/blob/main/dev-requirements.txt#L4

sunishsheth2009:
@mengxr Yes I thought about changing that: https://github.com/databricks/mlops-stack/blob/main/dev-requirements.txt#L4
but then realized that the code linked is not part of MLP. so it could download another version of MLFlow independent of the pipeline. So didn't change that. Let me know if I am missing something. 
cc @vladimirk-db and @smurching

sunishsheth2009:
> @sunishsheth2009 thanks for the PR, looks good! Do we plan to update the example ML code in the stacks to use MLflow 2.0 once it's out?

Yes we have a follow up task to update that repo with the new 2.0 changes. :)

PR #18, opened 2022-11-04
[ML-26476]Rename mlflow pipeline to mlflow recipe

mingyu89:
Rename mlflow pipeline to mlflow recipe.
 - Update any code that uses mlflow pipeline
   - only mlflow pipeline is renamed. Azure pipeline, ML pipeline and sklearn pipeline are not touched. 
 - Update requirements files to install mlflow hash that has the renaming commit
   - TODO: update requirements files again once the renaming is in a release
 - Update docs to use mlflow recipe
   - Any link to mlflow pipeline doc are not modified
   - TODO: update doc links once mlflow recipes is in a release
 - Rename pipeline.yaml to recipe.yaml

Unit Test 
Screen Shot 2022-11-04 at 3 42 36 PM


README.md
```diff
@@ -7,7 +7,7 @@ The default stack in this repo includes three modular components:
...
 |----------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| ML Code                    | Example ML project structure, with unit tested Python modules and notebooks using [MLflow pipelines](https://mlflow.org/docs/latest/pipelines.html)                   | Quickly iterate on ML problems, without worrying about refactoring your code into tested modules for productionization later on.                                                        |
+| ML Code                    | Example ML project structure, with unit tested Python modules and notebooks using [MLflow recipes](https://mlflow.org/docs/latest/pipelines.html)                     | Quickly iterate on ML problems, without worrying about refactoring your code into tested modules for productionization later on.                                                        |
```
sunishsheth2009, line 10: We can probably change the link `https://mlflow.org/docs/latest/pipelines.html` already since I am assuming that before merging we could have the new URLs deployed already. What do you think?
This implies to all the links in this PR

{{cookiecutter.project_name}}/notebooks/Train.py
```diff
@@ -58,9 +58,9 @@
 
 # COMMAND ----------
 
-from mlflow.pipelines import Pipeline
+from mlflow.recipes import Recipe
 
-p = Pipeline(profile=profile)
+p = Recipe(profile=profile)
```
sunishsheth2009, line 63: Can we change this to `r = Recipe(profile=profile)`

{{cookiecutter.project_name}}/recipe.yaml
```diff
@@ -85,5 +85,5 @@ metrics:
       function: weighted_mean_squared_error
       greater_is_better: False
   # Sets the primary metric to use to evaluate model performance. This primary metric is used
-  # to sort MLflow Runs corresponding to the pipeline in the MLflow Tracking UI
+  # to sort MLflow Runs corresponding to the recipe in the MLflow Tracking UI
```
sunishsheth2009, line 88: Unfortunately I can't make the comment on other files but can we fix the recipe.yaml with the new product spec?

Example in recipe:

`metrics` are renamed to `custom_metrics`
We don't have `data_scoring` step anymore, etc

The new spec can we found here: https://github.com/mlflow/mlp-regression-example/pull/14/files#diff-1250fe52790be66c0718296d26be97fdaee7cfca04d0448f5b87dba907dfa669

README.md
mingyu89: The current PR uses pip to install a commit version from github. We can change the link and the requirement files together once it's released.

{{cookiecutter.project_name}}/notebooks/Train.py
mingyu89: updated

{{cookiecutter.project_name}}/recipe.yaml
mingyu89: It worth a different PR to re-copy the new recipe file once it's released.

mingyu89: Nevermind about the "released". I'll have a followup PR to address this.

dev-requirements.txt
```diff
@@ -1,4 +1,4 @@
 cookiecutter>=2.1.0
 pytest
 pytest-black
-mlflow[pipelines]>=1.29.0
+git+https://github.com/mlflow/mlflow.git@4b259e
```
jinzhang21, line 4: Let's change this to `mlflow>=2.0.0`. The pre-release version is already published: https://pypi.org/project/mlflow/2.0.0rc0/

README.md
jinzhang21: I'd suggest to make it self-consistent with mlflow 2.0 release. Tracking these small changes are time consuming and doesn't worth that much given the release is only 1 week away.

jinzhang21 (approved):
Synced offline. Now I understand why. Let's merge it as is then. But also file a ticket to track the change once MLflow 2.0 is released with rename changes.

mingyu89:
Followup task ML-26726

PR #19, opened 2022-11-07
Add Databricks Feature Store Component to MLOps Stacks

vladimirk-db:
This PR adds Databricks Feature Store support as a component in MLOps Stacks. It follows similar principles applied to training code support, including sample feature engineering code, automated unit and integration testing and infra-as-code setup for deployment.

It includes the following:
- Sample feature code (transform) modules.
- Parameterized ETL notebooks that load data, apply feature modules and write toe Feature Store tables.
- Infra as code terraform modules to deploy above ETL notebooks as jobs.
- Sample model training code that is integrated with feature store client libraries.
- Automated  (via Github Workflows) integration tests that trigger feature engineering pipelines and training code and post results to Github UI as PR comments.

The sample code is based on the Feature Store [Taxi Example Notebook](https://docs.databricks.com/_static/notebooks/machine-learning/feature-store-taxi-example.html).

This functionality is optional and guarded by a  cookiecutter boolean flag (`include_feature_store`).

Tested:
* Unit tests
* The integration test changes have been validated in a separate demo repo.
* Will instantiate and validate against a demo repo before merging.
  -   Sample diff: [regenerated MLOps stack](https://github.com/databricks/mlops-azure-cuj/pull/36/files) on Azure w/ GH Actions with Feature Store **Disabled**.
  - Sample diff: [regenerated MLOps stack](https://github.com/databricks/mlops-azure-cuj/pull/37/files/ba72405343003486d959d21785fa760635c998c4) on Azure w/ GH Actions with Feature Store **Enabled**.
  - In both cases all unit + integration tests pass.

TODOs (fixed):

- [x] Post-instantiation Github Actions ML code unit tests failing (when initializing a [local Spark session](https://github.com/databricks/mlops-stack/actions/runs/3412002486/jobs/5676934626)).
- [x] Need to update docs to remove MLP references when Feature Store component is added.


tests/test_create_project.py
```diff
@@ -235,7 +236,7 @@ def test_strip_slash_if_needed_from_mlflow_experiment_parent_dir(
 
 
 @parametrize_by_project_generation_params
-def test_generate_project_with_default_values(tmpdir, cloud, cicd_platform):
+def test_generate_project_with_default_values(tmpdir, cloud, cicd_platform, include_feature_store):
```
mohamad-arabi, line 239: How is this new param used in the below test?

{{cookiecutter.project_name}}/.github/workflows/run-tests-fs.yml
```diff
@@ -0,0 +1,146 @@
...
+                    "env": "staging",
+                    "test_mode": "True",
+                    "input_table_path": "/databricks-datasets/nyctaxi-with-zipcodes/subsampled",
+                    "timestamp_column": "tpep_pickup_datetime",
+                    "output_table_name": "feature_store_taxi_example.trip_pickup_features-test",
+                    "features_transform_module": "pickup_features",
+                    "primary_keys": "zip"
+                  }
+                },
+                "existing_cluster_id": "0729-223229-zfjw22fr"
```
mohamad-arabi, line 66: Which cluster is this? We can't assume a cluster with this id exists everywhere the mlops stack is instantiated

{{cookiecutter.project_name}}/.github/workflows/run-tests-fs.yml
```diff
@@ -0,0 +1,146 @@
...
+            }`
+            return output.replace(/\r?\n|\r/g, '')    
+      - name: Feature Store/Model Training Integration Test
+        id: features-training-integration-test
+        run: |
+          python -m pip install --upgrade pip
+          pip install databricks-cli
+          databricks jobs configure --version=2.1
+          echo {% raw %}${{steps.integration-test-content.outputs.result}} > test.json {% endraw %}
+          databricks runs submit --json-file test.json --wait > tmp-output.json
```
mohamad-arabi, line 127: Isn't better if we chain a couple of `databricks/run-notebook@v0` together instead of in our stack the use of both the github action and the `databricks runs submit --wait` cli command.

Something more like this: 
```
 - name: Run first notebook
        uses: databricks/run-notebook@v0
        id: notebook-1
        with:
          local-notebook-path: path/to/first/notebooks_1.py
          ... // more fields to populate
          notebook-params-json: >
            {
              "input_1": "Hello",
              "input_2": "World!"
            }
          timeout-seconds: 1000
      - name: Run second notebook
        uses: databricks/run-notebook@v0
        id: notebook-2
        with:
          local-notebook-path: path/to/second/notebooks_2.py
          ... // more fields to populate
          notebook-params-json: >
            {
              "input_from_previous_notebook": "${{steps.notebook-1.outputs.notebook-output}}"
            }
          timeout-seconds: 1000
```

{{cookiecutter.project_name}}/requirements.txt
```diff
@@ -4,3 +4,6 @@ pandas>=1.4.3
 scikit-learn>=1.1.1
 matplotlib>=3.5.2
 Jinja2==3.0.3
+pyspark~=3.3.0
+pytz~=2022.2.1
+requests~=2.28.1
```
mohamad-arabi, line 9: What is this used for? I don't see `requests` imported anywhere

mohamad-arabi (commented):
Added some comments from first pass, Thanks Vlad

hooks/post_gen_project.py
```diff
@@ -12,6 +12,26 @@ def remove_filepath(filepath):
...
+    os.path.join("tests", "ingest_test.py"),
+    os.path.join("tests", "split_test.py"),
+    os.path.join("tests", "train_test.py"),
+    os.path.join("tests", "test_sample.parquet"),
+    os.path.join("tests", "transform_test.py")
+]
+
+feature_store_paths = [
+    os.path.join("features"),
+    os.path.join("databrics", "TrainWithFeatureStore.py"),
```
smurching, line 30:
```suggestion
    os.path.join("databricks", "TrainWithFeatureStore.py"),
```

cookiecutter.json
```diff
@@ -8,5 +8,6 @@
   "default_branch": "Name of the default branch, where the prod and staging ML resources are deployed from and the latest ML code is staged. Default: `main`",
   "release_branch": "Name of the release branch. The production jobs (model training, batch inference) defined in this stack pull ML code from this branch. Default: `release`",
   "read_user_group": "Display name of the user group to give read permissions to the ML jobs, integration test job runs, and machine learning resources created for this project. A group with this name must exist in both the staging and prod workspaces. Default: `users`",
+  "include_feature_store": false,
```
smurching, line 11: It would be nice to have this parameter documented like the others - but requires some logic to replace the default value with the correct default & validate user input (since we'd need to make this a string parameter)

{{cookiecutter.project_name}}/.github/workflows/run-tests-fs.yml
```diff
@@ -0,0 +1,183 @@
...
+        uses: actions/checkout@v3
+        {%- if cookiecutter.cloud == "azure" %}
+      - name: Generate AAD Token
+        {% raw %}run: ./.github/workflows/scripts/generate-aad-token.sh ${{ secrets.STAGING_AZURE_SP_TENANT_ID }} ${{ secrets.STAGING_AZURE_SP_APPLICATION_ID }} ${{ secrets.STAGING_AZURE_SP_CLIENT_SECRET }}{% endraw %}
+        {%- endif %}
+      - name: Fetch integration test
+        uses: actions/github-script@v6
+        id: integration-test-content
+        with:
+          # TODO update the tasks and notebook parameters below to match your integration test setup.
```
smurching, line 47: I wonder if we should document what this doing a bit in code comments, namely that we're building up the JSON payload to trigger the job run - it's a little tricky to infer from the step name "Fetch integration test"

{{cookiecutter.project_name}}/.github/workflows/run-tests-fs.yml
smurching: Yeah, this option is fine for now, but would be good to prioritize the change @vladimirk-db  mentioned if we find that people are regularly working with this file - the GitHub workflow contents are a little tricky to modify as it stands (a combination of GHA parameter substitution, Javascript, and JSON contents)

{{cookiecutter.project_name}}/.github/workflows/run-tests-fs.yml
vladimirk-db: Added more information (also renamed the name of the step).

cookiecutter.json
vladimirk-db: I changed this to a choice variable in the meantime (as boolean support is not official yet.) . Is there no way to document these types of variable, hmm looking into it.

{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
```diff
@@ -13,7 +13,7 @@ resource "databricks_job" "model_training_job" {
   task {
     task_key = "Train"
 
-    {% if cookiecutter.include_feature_store %}notebook_task {
+    {% if cookiecutter.include_feature_store == "yes" %}notebook_task {
```
smurching, line 16: Nit: we could actually in the `pre_gen_project.py` hook convert the "yes" or "no" input to this parameter into a boolean, so that we can just do the boolean check in project files

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
```diff
@@ -20,6 +20,12 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.8
+       {%- if cookiecutter.include_feature_store == "yes" %}
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'temurin'
+          java-version: '11'
```
smurching, line 27: I missed this originally, out of curiosity why do the feature store tests require Java? Should we be documenting the Java requirement in the ML developer guide for feature store e.g. in the prereqs for developing locally?

smurching (commented):
Basically LGTM, just a few last comments!

cookiecutter.json
smurching: Ah I see, yeah I guess no way to document choice variables, this is fine as is / clear enough I think - the alternative would have been to have a text variable where we document + validate that the user-supplied value is either "yes" or "no", which is arguably not much clearer

{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
vladimirk-db: Good idea, done: https://github.com/databricks/mlops-stack/pull/19/commits/a84cc7927c2a81b585899089118c28a20eb2c81f

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
vladimirk-db: Updated. Java is needed as we are creating a local Spark session for testing.

{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
vladimirk-db: Hmm, this doesn't completely work, so I reverted the commit. It gets a bit messy and I think the main reason is that this version of cookiecutter doesn't support boolean variables. Let's clean this up once we upgrade cookiecutter to 2.2.0 (w/ proper boolean vars).

smurching (approved):
LGTM!

PR #23, opened 2022-11-09
Tag clusters used for ADO CI/CD workflows

smurching:
Tag clusters used for Azure DevOps CI/CD workflows as originating from MLOps stacks, to help categorize clusters/attribute cost back to MLOps stacks workloads


mohamad-arabi (approved):
LGTM! should we add a ADO vs Gtihub Actions specific tags on the clusters. So that we can track them differently? This could be a seperate PR

PR #26, opened 2022-11-15
Add Feature Store Readme.

vladimirk-db:


README.md
```diff
@@ -109,14 +110,15 @@ best practices. See [the stack customization guide](stack-customization.md)
 for details on how to do this.
 
 ### Does the MLOps stack cover data (ETL) pipelines?
+
 While some components of the MLOps stack can be re-used and customized for ETL pipelines,
-the focus of the Databricks MLOps stack is on productionizing ML workflows: training, model deployment, batch inference, etc.
+the focus of the Databricks MLOps stack is on productionizing ML workflows: feature engineering, training, model deployment, batch inference, etc.
```
smurching, line 115: Curious what we actually recommend now that we have feature store support - would we expect people to use DLT or DLT meta to ETL some silver tables, and then use the feature store APIs to build final gold model training tables? Maybe we can double check with feature store team on what our story should be - I couldn't find a specific recommendation on how the two fit together in [Databricks docs](https://docs.databricks.com/machine-learning/feature-store/feature-tables.html#create-a-feature-table-in-databricks-feature-store)

vladimirk-db: Agreed, I'll follow up with Feature Store team and provide recommendations.

PR #36, opened 2022-12-09
Change Azure DevOp pipelines to be run automatically on a repo level

jerrylian-db:
## Description

Before this PR, ADO pipelines created by MLOps Stacks for one repo in an ADO project would be run on pull requests against all other repos in the same ADO project. This PR isolates the ADO branch policies on a repo level. It also augments the names of provisioned things with the repo name for clarity.

## Testing

I created two MLOps Stack projects based off of this feature branch and ran through the bootstrap scripts for both of them as two repos in the same ADO project. This resulted in the following ADO pipelines:
Screen Shot 2022-12-12 at 3 28 40 PM

Then, I created one ML code PR and one Databricks Config PR for one of the repos. See below that only the ADO pipelines for that repo were triggered:

Screen Shot 2022-12-12 at 3 29 49 PM

Screen Shot 2022-12-12 at 3 30 13 PM

### DIFF
--- a/{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/azure-devops.tf
+++ b/{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/azure-devops.tf
@@ -41,7 +41,7 @@ resource "azurerm_role_assignment" "prod_role_assignment" {
 // Create variable group to be used by Azure DevOps Pipelines
 resource "azuredevops_variable_group" "cicd_vg" {
   project_id   = data.azuredevops_project.project.project_id
-  name         = "CICD Variable Group"
+  name         = "CICD Variable Group [${data.azuredevops_git_repository.repo.name}]"
   description  = "Variable group for CICD pipelines"
   allow_access = true
 
@@ -106,7 +106,7 @@ resource "azuredevops_variable_group" "cicd_vg" {
 
 resource "azuredevops_build_definition" "testing-ci" {
   project_id = data.azuredevops_project.project.project_id
-  name       = "testing_ci"
+  name       = "testing_ci_${data.azuredevops_git_repository.repo.name}"
 
   ci_trigger {
     use_yaml = true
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/azure-devops.tf
```diff
@@ -160,10 +160,6 @@ resource "azuredevops_branch_policy_build_validation" "testing-ci-build-validati
       repository_ref = data.azuredevops_git_repository.repo.default_branch
       match_type     = "Exact"
     }
-
-    scope {
-      match_type = "DefaultBranch"
```
mohamad-arabi, line 165: Can you please add a comment as to why this is necessary/why it fixed the issue.

mohamad-arabi (approved):
LGTM, if small comment is addressed

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/azure-devops.tf
jerrylian-db: I scoured the internet in search of any RESP API documentation for what the different ways to configure `scope` mean exactly. Unfortunately, I couldn't find any concrete documentation. The best I could find was here: https://registry.terraform.io/providers/microsoft/azuredevops/latest/docs/resources/branch_policy_build_validation#argument-reference

Here was my thinking process making this change: from the terraform doc, it seems like users can configure multiple scopes. My assumption is that these scopes are clauses for where to configure this branch policy. The existing `Exact` scope seems to already accomplish what we want, which is matching to the default branch of the user specified repo. Then, with regards to the `DefaultBranch` scope, the terraform doc specified that users cannot configure `repository_id` nor `repositor_ref` when the `DefaultBranch` match type is chosen. My interpretation here is that that means that the `DefaultBranch` match type configures the policy on a project level, which is coherent my dogfooding experience of the old code.

Hence, I feel pretty confident that removing this scope is the fix that we need.

PR #42, opened 2023-02-01
[ML-27304][WIP] Remove unused env params in Feature Store ETL and Training notebooks

zhe-db:
Removed unused env params in Feature Store ETL and Training notebooks

- {{cookiecutter.project_name}}/notebooks/GenerateAndWriteFeatures.py
- {{cookiecutter.project_name}}/notebooks/TrainWithFeatureStore.py
- And upstream TF files and CICD Yml files

The reason we remove the env param is because

- Staging and production models are already differentiated by their name in the TF config files: https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/prod/training-job.tf#L22
- There is no need for differentiating feature tables by env and there is no use of env in downsteam APIs that we called.

Additional manual tests:
- Ran the notebooks on Databricks.
- Setup a dummy project on GitHub and created a dummy PR that triggers CI.

Testing with MLOps Azure CUJ Repo: https://github.com/databricks/mlops-azure-cuj/pull/55

### DIFF
--- a/{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
+++ b/{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
@@ -95,7 +95,6 @@ jobs:
                 "notebook_task": {
                   "notebook_path": "notebooks/Train",
                   "base_parameters": {
-                    "env": "staging",
                     "test_mode": "True"
                     }
                   },
--- a/{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
+++ b/{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
@@ -95,6 +95,7 @@ jobs:
                 "notebook_task": {
                   "notebook_path": "notebooks/Train",
                   "base_parameters": {
+                    "env": "staging"
                     "test_mode": "True"
                     }
                   },
--- a/{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
+++ b/{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
@@ -95,7 +95,7 @@ jobs:
                 "notebook_task": {
                   "notebook_path": "notebooks/Train",
                   "base_parameters": {
-                    "env": "staging"
+                    "env": "staging",
                     "test_mode": "True"
                     }
                   },
--- a/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
+++ b/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
@@ -79,7 +79,6 @@ jobs:
           run-name: {{ cookiecutter.project_name}} Integration Test
           notebook-params-json: >
             {
-              "env": "staging",
               "test_mode": "True"
             }
           {% raw %}pr-comment-github-token: ${{ secrets.GITHUB_TOKEN }}{% endraw %}
--- a/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
+++ b/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
@@ -77,7 +77,7 @@ jobs:
               }
             ]
           run-name: {{ cookiecutter.project_name}} Integration Test
-          {%- if cookiecutter.include_feature_store != "yes" %}
+          {%- if cookiecutter.include_feature_store == "no" %}
           notebook-params-json: >
             {
               "env": "staging",
--- a/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
+++ b/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
@@ -77,8 +77,17 @@ jobs:
               }
             ]
           run-name: {{ cookiecutter.project_name}} Integration Test
+          {%- if cookiecutter.include_feature_store == "yes" %}
           notebook-params-json: >
             {
               "test_mode": "True"
             }
+          {%- else %}
+          notebook-params-json: >
+            {
+              "env": "staging",
+              "test_mode": "True"
+            }
+          {%- endif %}
+          
           {% raw %}pr-comment-github-token: ${{ secrets.GITHUB_TOKEN }}{% endraw %}
--- a/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
+++ b/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
@@ -77,11 +77,7 @@ jobs:
               }
             ]
           run-name: {{ cookiecutter.project_name}} Integration Test
-          {%- if cookiecutter.include_feature_store == "yes" %}
-          notebook-params-json: >
-            {
-            }
-          {%- else %}
+          {%- if cookiecutter.include_feature_store != "yes" %}
           notebook-params-json: >
             {
               "env": "staging",
--- a/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
+++ b/{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
@@ -80,7 +80,6 @@ jobs:
           {%- if cookiecutter.include_feature_store == "yes" %}
           notebook-params-json: >
             {
-              "test_mode": "True"
             }
           {%- else %}
           notebook-params-json: >
--- a/{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
+++ b/{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
@@ -24,6 +24,7 @@ resource "databricks_job" "model_training_job" {
     {%- else -%}notebook_task {
       notebook_path = "notebooks/Train"
       base_parameters = {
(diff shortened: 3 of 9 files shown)


arpitjasa-db (commented):
Excellent work on this @zhe-db! Could we add in the description a bit more about why we're making this change instead of plumbing the params through for example? And it would be good to do a bit of testing to confirm the notebooks can still run as expected

arpitjasa-db:
Also don't forget to change [write-feature-table-job.tf](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/prod/write-feature-table-job.tf) and [training-job.tf](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/prod/training-job.tf) in both `staging` and `prod` to not pass in the `env` variables

zhe-db:
> Also don't forget to change [write-feature-table-job.tf](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/prod/write-feature-table-job.tf) and [training-job.tf](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/prod/training-job.tf) in both `staging` and `prod` to not pass in the `env` variables

Done, also remove the occurrences in other tf and yaml files.

{{cookiecutter.project_name}}/databricks-config/staging/training-job.tf
```diff
@@ -25,7 +24,6 @@ resource "databricks_job" "model_training_job" {
     {%- else -%}notebook_task {
       notebook_path = "notebooks/Train"
       base_parameters = {
-        env = local.env
```
mingyu89, line 28: For notebooke Train, the env is needed.

{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
```diff
@@ -25,7 +24,6 @@ resource "databricks_job" "model_training_job" {
     {%- else -%}notebook_task {
       notebook_path = "notebooks/Train"
       base_parameters = {
-        env = local.env
```
mingyu89, line 28: For notebooke Train, the env is needed.

mingyu89 (changes requested):

{{cookiecutter.project_name}}/.azure/devops-pipelines/tests-ci.yml
```diff
@@ -95,7 +95,6 @@ jobs:
                 "notebook_task": {
                   "notebook_path": "notebooks/Train",
                   "base_parameters": {
-                    "env": "staging",
```
mingyu89, line 98: Ditto

{{cookiecutter.project_name}}/databricks-config/staging/training-job.tf
zhe-db: thanks for the catch Mingyu, fixed.

mingyu89 (commented):
LGTM. I'll leave it to @arpitjasa-db to take another look and approve.

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
```diff
@@ -77,9 +77,17 @@ jobs:
               }
             ]
           run-name: {{ cookiecutter.project_name}} Integration Test
+          {%- if cookiecutter.include_feature_store == "yes" %}
```
arpitjasa-db, line 80: Wait why are we enabling `test_mode` by default for feature store?

zhe-db: yeah, `test_mode` is not needed, the payload for feature store ones should be empty

arpitjasa-db: Wait why do we need this at all? Can we revert?

arpitjasa-db: It looks weird to just have notebook params be empty. Can we rewrite this so the params are only added if it doesn't include feature store?

zhe-db: Because if [cookiecutter.include_feature_store is true, we set the local-notebook-path to be `otebooks/TrainWithFeatureStore.py`](https://github.com/databricks/mlops-stack/blob/3f9251ce8c22379852405f5d55dc6716dd7c71a8/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/run-tests.yml#L52), which does not take `env` nor `test_mode`?

zhe-db: cool

> It looks weird to just have notebook params be empty. Can we rewrite this so the params are only added if it doesn't include feature store?

cool

{{cookiecutter.project_name}}/databricks-config/staging/training-job.tf
```diff
@@ -16,7 +16,6 @@ resource "databricks_job" "model_training_job" {
     {% if cookiecutter.include_feature_store == "yes" %}notebook_task {
       notebook_path = "notebooks/TrainWithFeatureStore"
       base_parameters = {
-        env                = local.env
         training_data_path = "/databricks-datasets/nyctaxi-with-zipcodes/subsampled"
         experiment_name    = databricks_mlflow_experiment.experiment.name
         model_name         = "${local.env_prefix}{{cookiecutter.model_name}}"
```
vladimirk-db, line 21: Nitpick and note for future consideration: the environment value is used implicitly in this notebook. It's great that we are cleaning up the explicit unused param env, we should audit and figure out how to make "env" naming consistent across these notebooks in the future. For instance, models have environment prefixed in their name, but feature store tables don't.

vladimirk-db (approved):
Nicely done, @zhe-db !  LGTM

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
```diff
@@ -77,9 +77,12 @@ jobs:
               }
             ]
           run-name: {{ cookiecutter.project_name}} Integration Test
+          {%- if cookiecutter.include_feature_store != "yes" %}
```
arpitjasa-db, line 80: Nit: But what if we make this `== "no"` instead? Bit cleaner code and easier for our customers to understand

mingyu89 (approved):

arpitjasa-db (approved):
Looks great thanks for all the quick fixes @zhe-db! Let's confirm it all works though before merging

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
```diff
@@ -77,9 +77,12 @@ jobs:
               }
             ]
           run-name: {{ cookiecutter.project_name}} Integration Test
+          {%- if cookiecutter.include_feature_store == "no" %}
```
mingyu89, line 80: @arpitjasa-db It's fine as long as the condition is consistent across all places. Since other places are using `cookiecutter.include_feature_store == "yes"`, We should use `cookiecutter.include_feature_store != "yes"` here.

mingyu89 (changes requested):

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
arpitjasa-db: Wait @mingyu89 I disagree. In other places in stacks like [here](https://src.dev.databricks.com/databricks/mlops-stack/-/blob/%7B%7Bcookiecutter.project_name%7D%7D/docs/mlops-setup.md?L46:21), we use `"no"`. I think it would be better to stay consistent with this convention.

mingyu89: @arpitjasa-db  Ok. Then I'm Ok with fixing it in another PR. When we only check against one string, we are sure the condition is consistent for all places.

mingyu89 (approved):

{{cookiecutter.project_name}}/.github/workflows/run-tests.yml
mingyu89: It's not likely, but for example, if the value is "Yes" or "true", we treat it as "no feature store" across all places.(easier for user to figure out the outcome and the reason)

arpitjasa-db: Yeah I see your point, maybe we should do a better job instead of enforcing the schema described [here](https://src.dev.databricks.com/databricks/mlops-stack/-/blob/cookiecutter.json?L11:4).

mingyu89: @arpitjasa-db  We can use this once 2.2.0 is available. https://cookiecutter.readthedocs.io/en/latest/advanced/boolean_variables.html

PR #43, opened 2023-02-07
[ML-26956] Add support for Github Enterprise as CI/CD provider

zhe-db:
Added support for GitHub Enterprise as CI/CI provider

- Add an option for CICD provider when users create a project, now `GItHub Actions`, `GitHub Enterprise Actions`, `AzureDevOps`.
- Change `git-provider` to `gitHubEnterprise` payload for run-notebook github action according to [here](https://github.com/databricks/run-notebook/pull/29).
- Change `git-provider` payload to `gitHubEnterprise` for Databricks CLI to run notebooks according to  [here](https://src.dev.databricks.com/databricks/universe/-/blob/jobs-api/api/openapi/jobs-2.1.openapi.jsonnet?L1204).
- Change CD URL to `http(s)://HOSTNAME/api/v3/repos/OWNER/REPO/actions/workflows/WORKFLOW_ID/dispatches ` for gitHubEnterprise according to user supplied git repo URL according to github [docs](https://docs.github.com/en/enterprise-server@3.7/rest/actions/workflows).
- Add and fix unit tests.
- Testing with `mlops-azure-cuj`: https://github.com/databricks/mlops-azure-cuj/pull/60
- Testing workflow https://adb-3253686410953403.3.azuredatabricks.net/?o=3253686410953403#job/874940478032607/run/1231301


{{cookiecutter.project_name}}/notebooks/TriggerModelDeploy.py
```diff
@@ -45,15 +45,24 @@
...
 )
-cd_trigger_url = f"https://api.github.com/repos/{github_repo}/actions/workflows/deploy-model-{env}.yml/dispatches"
+if github_server == "https://github.com" or github_server == "github.com":
+    github_server = "https://api.github.com"
+else:
+    # URL for github Enterprise Server: https://docs.github.com/en/enterprise-server@3.8/rest/actions/workflows
+    github_server = f"{github_server}/api/v3"
+cd_trigger_url = f"{github_server}/repos/{github_repo}/actions/workflows/deploy-model-{env}.yml/dispatches"
 authorization = f"token {token}"
+print(cd_trigger_url)
```
zhe-db, line 65: used it for testing for now, will remove

README.md
```diff
@@ -50,7 +50,7 @@ To create a new project, run:
 This will prompt for parameters for project initialization. Some of these parameters are required to get started:
  * ``project_name``: name of the current project
  * ``cloud``: Cloud provider you use with Databricks (AWS, Azure, or GCP)
- * ``cicd_platform`` : CI/CD platform of choice (GitHub Actions or Azure DevOps)
+ * ``cicd_platform`` : CI/CD platform of choice (GitHub Actions or Github Enterprise Actions or Azure DevOps)
```
mingyu89, line 53: How about (Github Actions, Github Actions for enterprises or Azure DevOps)? 
The "cicd_platform" variable needs to be changed accordingly.

I wasn't able to find term "Github Enterprise Actions".
Github doc
https://docs.github.com/en/enterprise-server@3.8/admin/github-actions/getting-started-with-github-actions-for-your-enterprise/about-github-actions-for-enterprises

hooks/post_gen_project.py
```diff
@@ -55,7 +55,7 @@ def remove_filepath(filepath):
 
 
 cicd_specific_paths = {
-    "gitHub": [
+    "git": [
```
mingyu89, line 58: Could you explain purpose of the changes in this file?

{{cookiecutter.project_name}}/.mlops-setup-scripts/README.md
```diff
@@ -5,7 +5,9 @@ This directory contains setup scripts intended to automate CI/CD and ML resource
 for MLOps engineers.
 
 {% if cookiecutter.cicd_platform == "gitHub" -%}
-The scripts set up CI/CD with GitHub Actions. If using another CI/CD provider, you can
+The scripts set up CI/CD with GitHub Actions. 
+{% elif cookiecutter.cicd_platform == "gitHubEnterprise" -%}
```
mingyu89, line 9: Lines 10 to 20 are previous under "gitHub". After the change, they will be only under "gitHubEnterprise" and missing from "gitHub".

hooks/post_gen_project.py
zhe-db: `gitHub` and `gitHubEnterprise` both uses `.github` directory, that python cannot hash lists or sets as keys so I make a common key for both `gitHub` and `gitHubEnterprise`, the reason not to hash both `gitHub` and `gitHubEnterprise` is because of the below code blocks where we remove file paths of unwanted directories.

mingyu89: Got it. It's the unrelated cleanup logic after the project generation. You are matching the prefix to map `gitHub` and `gitHubEnterprise` to the same entity in the map.

Since there are only 2 options, how about Removing the map and explicitly checking the condition
`cicd_platform == "gitHub"/"gitHubEnterprise"/"azureDevOpsServices"` and remove related files?

In this way there's no implicit cicd_platform name assumption.

zhe-db: Agree, the remove_map logic is kinda weird, I will change to explicitly check condition.

hooks/post_gen_project.py
```diff
@@ -53,21 +53,11 @@ def remove_filepath(filepath):
...
-        os.path.join(".azure"),
-        os.path.join(".mlops-setup-scripts", "cicd", "azure-devops.tf"),
-    ],
-}
-
-for cicd, paths in cicd_specific_paths.items():
-    if cicd != cicd_platform:
-        for path in paths:
-            remove_filepath(path)
+if cicd_platform in ["gitHub", "gitHubeEnterprise"]:
```
arpitjasa-db, line 56:
```suggestion
if cicd_platform in ["gitHub", "gitHubEnterprise"]:
```

{{cookiecutter.project_name}}/.mlops-setup-scripts/README.md
```diff
@@ -198,6 +200,9 @@ python .mlops-setup-scripts/cicd/bootstrap.py \
 {%- if cookiecutter.cicd_platform == "gitHub" %}
   --var github_repo_url=https://github.com/<your-org>/<your-repo-name> \
   --var git_token=<your-git-token>
+{%- elif cookiecutter.cicd_platform == "gitHubEnterprise" %}
+   -- var github_repo_url=<your-enterprise-github-server>/<your-org>/<your-repo-name> \
+   -- var git_token=<your-git-token>
```
arpitjasa-db, line 205:
```suggestion
   --var github_repo_url=<your-enterprise-github-server>/<your-org>/<your-repo-name> \
   --var git_token=<your-git-token>
```
Also would it be `github_repo_url` or `github_enterprise_repo_url`?

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
```diff
@@ -19,10 +19,18 @@ resource "databricks_secret" "cd_credentials_token" {
   scope        = databricks_secret_scope.cd_credentials.id
 }
 
+resource "databricks_secret" "cd_github_server" {
+  key = "github_server"
+  // regex expression for parsing github server url out of github repo url
+  // example: http(s)://mygitserver.anything/org/repo -> http(s)://mygitserver.anything
+  string_value = regex(".*\/\/[^\/]+", var.github_repo_url)
+  scope = databricks_secrete_scope.cd_credentials.id
```
arpitjasa-db, line 27:
```suggestion
  scope = databricks_secret_scope.cd_credentials.id
```

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
```diff
@@ -19,10 +19,18 @@ resource "databricks_secret" "cd_credentials_token" {
...
+}
+
 resource "databricks_secret" "cd_github_repo" {
   key = "github_repo"
   // Extract the 'organization/repo' substring used to identify the repo
-  string_value = replace(var.github_repo_url, "https://github.com/", "")
+  string_value = replace(var.github_repo_url, databricks_secret.cd_github_server, "")
```
arpitjasa-db, line 33: I think you'd need to do `databricks_secret.cd_github_server.value` but that also may not work for sensitivity. Could be easier to just pull out the regex logic into a Terraform `local` and pass it into both `resources`.

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
```diff
@@ -19,10 +19,18 @@ resource "databricks_secret" "cd_credentials_token" {
   scope        = databricks_secret_scope.cd_credentials.id
 }
 
+resource "databricks_secret" "cd_github_server" {
+  key = "github_server"
+  // regex expression for parsing github server url out of github repo url
+  // example: http(s)://mygitserver.anything/org/repo -> http(s)://mygitserver.anything
+  string_value = regex(".*\/\/[^\/]+", var.github_repo_url)
```
arpitjasa-db, line 26: This regex scares me a bit, can we confirm this works as expected via some tests? Wonder if it would be simpler to ask them to provide it

{{cookiecutter.project_name}}/notebooks/TriggerModelDeploy.py
```diff
@@ -45,15 +45,24 @@
...
+    f"{env}-{{cookiecutter.project_name}}-cd-credentials", "github_server"
+)
 github_repo = dbutils.secrets.get(
     f"{env}-{{cookiecutter.project_name}}-cd-credentials", "github_repo"
 )
 token = dbutils.secrets.get(
     f"{env}-{{cookiecutter.project_name}}-cd-credentials", "token"
 )
-cd_trigger_url = f"https://api.github.com/repos/{github_repo}/actions/workflows/deploy-model-{env}.yml/dispatches"
+if github_server == "https://github.com" or github_server == "github.com":
```
arpitjasa-db, line 58: Might be safer to use `cookiecutter.cicd_platform` here as well instead of bifurcating on URLs (someone could use a `http` or throw in a `www`)

arpitjasa-db (commented):
Awesome work @zhe-db! Left a few minor comments

{{cookiecutter.project_name}}/.mlops-setup-scripts/README.md
zhe-db: I use the same variable for `gitHub` and `gitHubEnterprise` to reduce complexity since we only allow one choice.

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
zhe-db: Cool, sounds good.

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
zhe-db: Cool, will add another variable for github_server

{{cookiecutter.project_name}}/.mlops-setup-scripts/README.md
```diff
@@ -198,6 +211,10 @@ python .mlops-setup-scripts/cicd/bootstrap.py \
 {%- if cookiecutter.cicd_platform == "gitHub" %}
   --var github_repo_url=https://github.com/<your-org>/<your-repo-name> \
   --var git_token=<your-git-token>
+{%- elif cookiecutter.cicd_platform == "gitHubEnterprise" %}
+   --var github_repo_url=<your-enterprise-github-server>/<your-org>/<your-repo-name> \
+   --var git_token=<your-git-token>
```
zhe-db, line 216: missing a `\`, will add

{{cookiecutter.project_name}}/.github/workflows/run-tests-fs.yml
```diff
@@ -148,7 +148,7 @@ jobs:
             ],
             "git_source": {
               "git_url": {% raw %}"${{ github.server_url }}/${{ github.repository }}",{% endraw %}
-              "git_provider": "gitHub",
+              "git_provider": "{{cookiecutter.cicd_platform}}",
```
vladimirk-db, line 151: Can we verify diffs for the gitHub option as well (that there's no regression).

You can do two separate commits of a generated Stacks project and then diff them.

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
```diff
@@ -19,10 +19,16 @@ resource "databricks_secret" "cd_credentials_token" {
   scope        = databricks_secret_scope.cd_credentials.id
 }
 
+resource "databricks_secret" "cd_github_server" {
```
vladimirk-db, line 22: Wait, is the cd_github_server a databricks secret? with a scope?  It's not a credential, right?

vladimirk-db: (Maybe I'm missing something.)

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
```diff
@@ -19,10 +19,16 @@ resource "databricks_secret" "cd_credentials_token" {
...
+  key = "github_server"
+  string_value = var.github_server_url
+  scope = databricks_secret_scope.cd_credentials.id
+}
+
 resource "databricks_secret" "cd_github_repo" {
```
vladimirk-db, line 28: Same thing here - why is the github_repo a databricks_secret?

README.md
vladimirk-db: I wouldn't do Github Enterprise Actions... "Github Actions (Enterprise)" or "Github Actions for Enterprises" maybe? The latter seems to be the official name.

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
zhe-db: I am following the same style for `github_repo` [here](https://src.dev.databricks.com/databricks/mlops-stack/-/blob/%7B%7Bcookiecutter.project_name%7D%7D/.mlops-setup-scripts/cicd/common/main.tf?L22). I think the reason is to facilitate retrieving the value through dbutils for example [here](https://src.dev.databricks.com/databricks/mlops-stack/-/blob/%7B%7Bcookiecutter.project_name%7D%7D/notebooks/TriggerModelDeploy.py?L49).

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
zhe-db: Please see above and correct me if I explained it wrong @arpitjasa-db

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
vladimirk-db: Ack, please double check this w/ @arpitjasa-db .

README.md
arpitjasa-db: Looks like the official name is [GitHub Actions for GitHub Enterprise Server](https://docs.github.com/en/enterprise-server@3.5/admin/github-actions/getting-started-with-github-actions-for-your-enterprise/getting-started-with-github-actions-for-github-enterprise-server)

{{cookiecutter.project_name}}/notebooks/TriggerModelDeploy.py
```diff
@@ -45,15 +45,27 @@
 assert env != "None", "env notebook parameter must be specified"
 assert model_uri != "", "model_uri notebook parameter must be specified"
 
-{% if cookiecutter.cicd_platform == "gitHub" -%}
+{% if cookiecutter.cicd_platform in ["gitHub", "gitHubEnterprise"] -%}
+github_server = dbutils.secrets.get(
```
arpitjasa-db, line 49: We can move this logic inside the `elif` on L61 since we only use the `github_server` value in the enterprise case

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/variables.tf
```diff
@@ -9,14 +9,20 @@ variable "git_provider" {
...
+  description = "URL of the hosted git server containing the current ML project, e.g. https://github.com/"
+  default     = "https://github.com/"
+}
+
 variable "env" {
   type        = string
   description = "Current env, i.e. 'staging' or 'prod'"
 }
-{%- endif -%}
+{%- endif -%}
```
arpitjasa-db, line 28: nit newline

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/common/main.tf
arpitjasa-db: Yeah we could consider instead writing out to a `json` output for example and pulling the value in from there for non-sensitive values, but this is fine for now since it's following the preset convention

vladimirk-db: The issue is that it's misleading - for instance the scope value does not apply, but it's present. Can we at least document this and then follow-up (don't need to block this PR on it).

README.md
vladimirk-db: Damn, that's a mouthful but it's better (can't argue with official).

vladimirk-db (approved):
LGTM @zhe-db !

PR #45, opened 2023-02-09
[easy] Remove hard-coded MLOps Azure CUJ experiment name and model name

mingyu89:
Signed-off-by: Mingyu Li <mingyu.li@databricks.com>

## Before
Default experiment name `/mlops-azure-cuj/mlops-azure-cuj-experiment-test`
Default model name `mlops-azure-cuj-model-test`

---

## After
Create project with name "my-next-project". The correct experiment name and model name is reflected in the project.
Screen Shot 2023-02-09 at 1 59 52 PM

### DIFF
--- a/{{cookiecutter.project_name}}/notebooks/TrainWithFeatureStore.py
+++ b/{{cookiecutter.project_name}}/notebooks/TrainWithFeatureStore.py
@@ -24,10 +24,10 @@
 dbutils.widgets.text("training_data_path", "/databricks-datasets/nyctaxi-with-zipcodes/subsampled", label="Path to the training data")
 
 # MLflow experiment name.
-dbutils.widgets.text("experiment_name", "/mlops-azure-cuj/mlops-azure-cuj-experiment-test", label="MLflow experiment name")
+dbutils.widgets.text("experiment_name", "/{{cookiecutter.project_name}}/{{cookiecutter.experiment_base_name}}-test", label="MLflow experiment name")
 
 # MLflow registered model name to use for the trained mode..
-dbutils.widgets.text("model_name", "mlops-azure-cuj-model-test", label="Model Name")
+dbutils.widgets.text("model_name", "{{cookiecutter.model_name}}-test", label="Model Name")
 
 # COMMAND ----------
 # DBTITLE 1,Define input and output variables


zhe-db (approved):

vladimirk-db (approved):
Thank you!

{{cookiecutter.project_name}}/notebooks/TrainWithFeatureStore.py
```diff
@@ -24,10 +24,10 @@
 dbutils.widgets.text("training_data_path", "/databricks-datasets/nyctaxi-with-zipcodes/subsampled", label="Path to the training data")
 
 # MLflow experiment name.
-dbutils.widgets.text("experiment_name", "/mlops-azure-cuj/mlops-azure-cuj-experiment-test", label="MLflow experiment name")
+dbutils.widgets.text("experiment_name", "/{{cookiecutter.project_name}}/{{cookiecutter.experiment_base_name}}-test", label="MLflow experiment name")
```
arpitjasa-db, line 27: Do we need the `-test` here? Same for model below? I think it's fine if we just take the customer's input and use that solely for the experiment/model name

mingyu89: Not 100% sure but I'm assuming if nothing is provided, it's test mode?

arpitjasa-db: Ah if we allow these fields to be blank, then it would be especially weird to allow starting a name with `-`

mingyu89: I mean if user sync the project to databricks repo and open the file as notebook and run it, I think that's the "test mode".
In this case the experiment_name and model name parameters will not be provided. 
We will use the default value for the experiment_name and model name, which will be derived from cookiecutter (`/mlops-azure-cuj/mlops-azure-cuj-experiment-test` and `mlops-azure-cuj-model-test` for the CUJ project).

arpitjasa-db: Hmm for [experiment](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/staging/mlflow-experiment.tf) and [model](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/databricks-config/staging/mlflow-model.tf) looks like we prefix with `env` anyways. But I think a suffix could make sense, but maybe something more descriptive than just `-test` since it seems confusing

mingyu89: @arpitjasa-db What's your suggestion? The test mode suffix is consistent with non-feature-store case.(in the mlflow recipes test profile)
We can modify both together.

Screen Shot 2023-02-09 at 4 45 49 PM

arpitjasa-db: Yeah but we only use the [test profile](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/profiles/databricks-test.yaml) in [Train.py](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/notebooks/Train.py) where we don't assume `-test` by default and for non-test mode, we take in the value from the other profiles, e.g. [staging profile](https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/profiles/databricks-staging.yaml). We don't have a test mode parameter for the feature store training notebook, so I'm just a bit confused why we're using this `-test` suffix by default for all env etc.

mingyu89: @arpitjasa-db I'll merge the PR for now. Let's talk next week and resolve it as a followup.
I actually feel the feature-store default value approach is better than a separate test_mode variable.

arpitjasa-db: Alright I added it to our weekly agenda

PR #50, opened 2023-02-15
Add timestamp column and model version column to batch inference output table

mingyu89:
The model version column and timestamp column are required by databricks model monitoring API.

---

Test
Make the same change to mlops-azure-cuj branch model-validation-bugbash1 and run BatchInference workflow
Screen Shot 2023-02-15 at 1 28 13 PM

---

The table schema
Screen Shot 2023-02-16 at 2 39 22 PM

---

Output data
Screen Shot 2023-02-16 at 2 39 30 PM

### DIFF
--- a/{{cookiecutter.project_name}}/notebooks/BatchInference.py
+++ b/{{cookiecutter.project_name}}/notebooks/BatchInference.py
@@ -52,13 +52,21 @@ model_uri = f"models:/{model_name}/{stage}"
 
 # Get model version from stage
 from mlflow import MlflowClient
+
 model_version_infos = MlflowClient().search_model_versions("name = '%s'" % model_name)
-model_version = next(version for version in model_version_infos if version.current_stage == stage).version
+model_version = next(
+    version for version in model_version_infos if version.current_stage == stage
+).version
+
+# Get datetime
+from datetime import datetime
+
+ts = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
 
 # COMMAND ----------
 
 # DBTITLE 1,Load model and run inference
 from predict import predict_batch
 
-predict_batch(spark, model_uri, input_table_name, output_table_name, model_version)
+predict_batch(spark, model_uri, input_table_name, output_table_name, model_version, ts)
 dbutils.notebook.exit(output_table_name)
--- a/{{cookiecutter.project_name}}/notebooks/BatchInference.py
+++ b/{{cookiecutter.project_name}}/notebooks/BatchInference.py
@@ -47,7 +47,13 @@ assert input_table_name != "", "input_table_name notebook parameter must be spec
 assert output_table_name != "", "output_table_name notebook parameter must be specified"
 
 model_name = get_model_name(env)
-model_uri = f"models:/{model_name}/{get_deployed_model_stage_for_env(env)}"
+stage = get_deployed_model_stage_for_env(env)
+model_uri = f"models:/{model_name}/{stage}"
+
+# Get model version from stage
+from mlflow import MlflowClient
+model_version_infos = MlflowClient().search_model_versions("name = '%s'" % model_name)
+model_version = next(version for version in model_version_infos if version.current_stage == stage).version
 
 # COMMAND ----------
 
--- a/{{cookiecutter.project_name}}/notebooks/BatchInference.py
+++ b/{{cookiecutter.project_name}}/notebooks/BatchInference.py
@@ -61,7 +61,7 @@ model_version = next(
 # Get datetime
 from datetime import datetime
 
-ts = datetime.now().strftime("%m/%d/%Y, %H:%M:%S")
+ts = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
 
 # COMMAND ----------
 
--- a/{{cookiecutter.project_name}}/notebooks/BatchInference.py
+++ b/{{cookiecutter.project_name}}/notebooks/BatchInference.py
@@ -54,9 +54,9 @@ model_uri = f"models:/{model_name}/{stage}"
 from mlflow import MlflowClient
 
 model_version_infos = MlflowClient().search_model_versions("name = '%s'" % model_name)
-model_version = next(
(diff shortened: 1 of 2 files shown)


{{cookiecutter.project_name}}/steps/predict.py
```diff
@@ -11,7 +13,13 @@ def predict_batch(spark_session, model_uri, input_table_name, output_table_name)
     predict = mlflow.pyfunc.spark_udf(
         spark_session, model_uri, result_type="string", env_manager="conda"
     )
-    output_df = table.withColumn("prediction", predict(struct(*table.columns)))
+    output_df = (
+        table.withColumn("prediction", predict(struct(*table.columns)))
+        .withColumn("model_version", lit(model_version))
+        .withColumn("ts", lit(ts))
+        .withColumn("model_version", col("model_version").cast("int"))
+        .withColumn("ts", to_timestamp("ts"))
```
mingyu89, line 21: Convert them to correct type so that the data monitoring don't need additional step to convert them.

zhe-db (approved):

vladimirk-db (approved):

{{cookiecutter.project_name}}/notebooks/BatchInference.py
```diff
@@ -47,12 +47,26 @@
...
 model_name = get_model_name(env)
-model_uri = f"models:/{model_name}/{get_deployed_model_stage_for_env(env)}"
+stage = get_deployed_model_stage_for_env(env)
+model_uri = f"models:/{model_name}/{stage}"
+
+# Get model version from stage
+from mlflow import MlflowClient
+
+model_version_infos = MlflowClient().search_model_versions("name = '%s'" % model_name)
+model_version = next(
```
arpitjasa-db, line 57: Curious why do we use `next` here?

{{cookiecutter.project_name}}/steps/predict.py
```diff
@@ -11,7 +13,13 @@ def predict_batch(spark_session, model_uri, input_table_name, output_table_name)
     predict = mlflow.pyfunc.spark_udf(
         spark_session, model_uri, result_type="string", env_manager="conda"
     )
-    output_df = table.withColumn("prediction", predict(struct(*table.columns)))
+    output_df = (
+        table.withColumn("prediction", predict(struct(*table.columns)))
+        .withColumn("model_version", lit(model_version))
+        .withColumn("ts", lit(ts))
```
arpitjasa-db, line 19: Is there a reason we're adding a column for both the literal value and the casted value? Do we expect the cast to fail? Or since we're using the same column key we expect the latter cast calls to overwrite?

{{cookiecutter.project_name}}/steps/predict.py
arpitjasa-db, line 21: Bit worried `ts` isn't descriptive enough. Does data monitoring want it to be called `ts`? If not I think we can expand it to call it like `batch_timestamp` 🙂

arpitjasa-db (commented):
Awesome work @mingyu89! Left a few comments

{{cookiecutter.project_name}}/notebooks/BatchInference.py
mingyu89: I searched and the only way I found to get the version number of `Production` is to search it. 
The search result is list and I'm using `next` to get the element from it, which is the model version.

arpitjasa-db: Ah I see, do we ever expect the size of this list to be > 1 (since multiple versions can be in the same stage), and if so, is it ok to just return the first version?

{{cookiecutter.project_name}}/steps/predict.py
mingyu89: `model_version` cast is necessary and I'll remove it. string works.

I'm not familiar enough with spark. Let me try `to_timestamp(lit(ts))` and see if it works.

{{cookiecutter.project_name}}/steps/predict.py
mingyu89: The data monitoring example in the API reference is using `ts`, which is short for timestamp. 
It could be from batch inference or online serving I think.

{{cookiecutter.project_name}}/notebooks/BatchInference.py
mingyu89: It shouldn't be more than 1 because we are filtering based on the stage. There should be only 1 Production model version and 1 Staging model version.
If this is in the middle of a model deployment and two versions are Production due to race condition, it's fine to use either of them for inference.

mingyu89: If alias is in place and there are two `prod` versions, I think the batch should be split into two and each version gets half. It will be a bigger change for inference.

arpitjasa-db: Ah well actually we don't disallow two model versions to be in the same stage as of now
Screen Shot 2023-02-15 at 6 25 32 PM
For example in this case, Version 2 and 3 are both in "Production" but the URI would only return version 3 since it's the latest in the stage.

{{cookiecutter.project_name}}/steps/predict.py
arpitjasa-db: @vperiyasamy any chance we can make this `ts` column a bit more descriptive in the example as well 😛

{{cookiecutter.project_name}}/notebooks/BatchInference.py
mingyu89: That's surprising. Ok I'll use `max` instead of `next`.
If we are always returning the latest one, we should update to return a random one. Otherwise there's no use of keeping the stage of the old version.

mingyu89:
@arpitjasa-db I addressed your comments and tested with 2 production versions. All look fine.

arpitjasa-db (approved):
LGTM thanks so much @mingyu89 !

PR #51, opened 2023-02-15
Specify default GitHub URL variable

arpitjasa-db:
Updated Terraform Modules as well to separate SPs and directories to prevent conflict in single workspace:
- https://github.com/databricks/terraform-databricks-mlops-aws-project
- https://github.com/databricks/terraform-databricks-mlops-azure-project-with-sp-creation
- https://github.com/databricks/terraform-databricks-mlops-azure-project-with-sp-linking

Tested via https://github.com/databricks/single-workspace

### DIFF
--- a/{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/variables.tf
+++ b/{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/variables.tf
@@ -43,6 +43,9 @@ variable "github_repo_url" {
 variable "github_server_url" {
   type        = string
   description = "URL of the hosted git server containing the current ML project, e.g. https://github.com/"
+  {% if cookiecutter.cicd_platform == "gitHub" -%}
+  default     = "https://github.com/"
+  {% endif -%}
   validation {
     condition     = length(var.github_server_url) > 0
     error_message = "The github_server_url variable cannot be empty"
--- a/{{cookiecutter.project_name}}/databricks-config/prod/outputs.tf
+++ b/{{cookiecutter.project_name}}/databricks-config/prod/outputs.tf
@@ -5,31 +5,31 @@
 
 # Batch inference job
 
-output "prod_{{cookiecutter.project_name}}_batch_inference_job_id" {
+output "{{cookiecutter.project_name}}_batch_inference_job_id" {
   value = databricks_job.batch_inference_job.id
 }
 
 
 # Mlflow experiment
 
-output "prod_{{cookiecutter.project_name}}_experiment_id" {
+output "{{cookiecutter.project_name}}_experiment_id" {
   value = databricks_mlflow_experiment.experiment.id
 }
 
-output "prod_{{cookiecutter.project_name}}_experiment_name" {
+output "{{cookiecutter.project_name}}_experiment_name" {
   value = databricks_mlflow_experiment.experiment.name
 }
 
 
 # Model Registry registered model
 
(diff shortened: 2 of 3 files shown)


{{cookiecutter.project_name}}/databricks-config/prod/outputs.tf
```diff
@@ -5,31 +5,31 @@
 
 # Batch inference job
 
-output "{{cookiecutter.project_name}}_batch_inference_job_id" {
+output "prod_{{cookiecutter.project_name}}_batch_inference_job_id" {
```
arpitjasa-db, line 8: Not fully sure if we want to add these env prefixes since they're put into `staging.json` and `prod.json` respectively, but it could help make it clearer. Thoughts?

mingyu89: If you were to change it, the mlflow recipe yaml profile for prod/staging also needs to be changed. They read the value from the terraform output files.
The model validation also read experiment name from the terraform outputs(but I will change the logic to directly read it from terraform, in this sprint).

I feel it not necessary to add hardcoded prefix. Same names with different location helped with code reuse.

arpitjasa-db: Yeah there's also a couple places in the GitHub and ADO workflows that would need to be changed as well but before going through the effort to change all that, wanted validation that this is the right approach

vladimirk-db: Agree with @mingyu89 , we probably don't need to do it, I think it's fine to rely on the prod/staging directories to disambiguate.

{{cookiecutter.project_name}}/.mlops-setup-scripts/cicd/variables.tf
```diff
@@ -43,6 +43,9 @@ variable "github_repo_url" {
 variable "github_server_url" {
   type        = string
   description = "URL of the hosted git server containing the current ML project, e.g. https://github.com/"
+  {% if cookiecutter.cicd_platform == "gitHub" -%}
+  default     = "https://github.com/"
+  {% endif -%}
```
vladimirk-db, line 48: What happens if we don't specify this?

arpitjasa-db: I was manually prompted to add it in, which is weird because we already fill the other two variables as part of the command execution we recommend in the docs.

{{cookiecutter.project_name}}/databricks-config/prod/outputs.tf
arpitjasa-db: Great I'll revert this part then.

vladimirk-db (approved):

mingyu89:
Could you add the jira issue number?

arpitjasa-db:
> Could you add the jira issue number?

I thought we only did this in universe since this is a public repo and we don't expose JIRA numbers in public repos (e.g. `mlflow`)?

mingyu89:
@arpitjasa-db  WOW I need to come back and modify all my PR title.
I thought it's nice to have the ticket link to PR automatically.

mingyu89 (approved):

arpitjasa-db:
> @arpitjasa-db WOW I need to come back and modify all my PR title. I thought it's nice to have the ticket link to PR automatically.

Yeah haha I totally understand

PR #53, opened 2023-02-18
Add experiment name to terraform workflow input

mingyu89:
Add experiment name to terraform workflow input so that both staging and prod workspace can get the correct experiment name.

## Test
Made the same changes to CUJ project, successfully run the training workflow with validation enabled.
Screen Shot 2023-02-17 at 10 41 28 PM
Screen Shot 2023-02-17 at 10 41 40 PM

### DIFF
--- a/{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
+++ b/{{cookiecutter.project_name}}/databricks-config/prod/training-job.tf
@@ -49,7 +49,7 @@ resource "databricks_job" "model_training_job" {
     notebook_task {
       notebook_path = "notebooks/ModelValidation"
       base_parameters = {
-        env = local.env
+        env             = local.env
         experiment_name = databricks_mlflow_experiment.experiment.name
         # Run mode for model validation. Possible values are :
         #   disabled : Do not run the model validation notebook.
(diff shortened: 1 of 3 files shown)


vladimirk-db (approved):

vladimirk-db:
Do we need to make these changes in the integration tests as well (github actions) ?

mingyu89:
@vladimirk-db The terraform is validated by the tests(and caught the format issue). The notebook is not covered by unit test, as it doesn't have pre-filled parameters and can not run on local system.

vladimirk-db:
> @vladimirk-db The terraform is validated by the tests(and caught the format issue). The notebook is not covered by unit test, as it doesn't have pre-filled parameters and can not run on local system.

I meant these integration tests that run on Databricks:
https://github.com/databricks/mlops-stack/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/run-tests.yml

Hmm I see, validation is not there as a step (should it be there? so we automatically do validation on changes in staging).

mingyu89:
@vladimirk-db Current integration test only covers training. If model validation is in disabled state, it may not align with user's expectation if CI/CD failed due to model validation failure/compiling error.

PR #55, opened 2023-02-22
Support directly running ModelValidation notebook in addition to workflow jobs

mingyu89:
## Support running ModelValidation notebook from workflow.
- Update the "env" parameter label to indicate it's used to decide which input data to use
- Read model name and model version from widgets if we can't get it from jobs util
- Fix model version description when there are multiple validations. 

**For moving the TODOs to a different python file, I have another ticket created.**

## Test running the model validation notebook several times
Screen Shot 2023-02-22 at 12 11 50 PM

#### The run
Screen Shot 2023-02-22 at 12 12 08 PM

#### model version description has a list of all validation runs
Screen Shot 2023-02-22 at 12 12 02 PM


{{cookiecutter.project_name}}/notebooks/ModelValidation.py
```diff
@@ -84,6 +87,11 @@ def get_targets_from_recipe():
 model_name = dbutils.jobs.taskValues.get("Train", "model_name", debugValue="")
 model_version = dbutils.jobs.taskValues.get("Train", "model_version", debugValue="")
 
+if model_uri == "" or model_name == "" or model_version == "":
```
arpitjasa-db, line 90: Is there ever a case where `model_uri` isn't empty but `model_name` or `model_version` are? And if so, do we still want to override the `model_uri`?

{{cookiecutter.project_name}}/notebooks/ModelValidation.py
```diff
@@ -150,10 +158,6 @@ def get_targets_from_recipe():
 
 # COMMAND ----------
 
-client.get_model_version(model_name, model_version).description
```
arpitjasa-db, line 153: Did this cell previously serve to output the description to the user? And if so, why did we decide to not do that?

{{cookiecutter.project_name}}/notebooks/ModelValidation.py
```diff
@@ -37,8 +37,11 @@
 
 # COMMAND ----------
 
-dbutils.widgets.dropdown("env", "prod", ["staging", "prod"], "Environment Name")
+dbutils.widgets.dropdown("env", "prod", ["staging", "prod"], "Environment Name(decide input data)")
```
arpitjasa-db, line 40: Yeah this is a tricky rename. Maybe something like this so it doesn't get cut off and the use makes sense.
```suggestion
dbutils.widgets.dropdown("env", "prod", ["staging", "prod"], "Environment (for input data)")
```

{{cookiecutter.project_name}}/notebooks/ModelValidation.py
mingyu89: No. All three will be set from the notebook of the previous job.

{{cookiecutter.project_name}}/notebooks/ModelValidation.py
mingyu89: I'm almost sure this is a command that I used for debugging and forgot to delete. It prints the current model version description, which the user doesn't care.

arpitjasa-db: Fantastic! Let's kill it 🙂

{{cookiecutter.project_name}}/notebooks/ModelValidation.py
mingyu89: Sure. cut-off content will be shown in a hover card.

{{cookiecutter.project_name}}/notebooks/ModelValidation.py
arpitjasa-db: So in that case it might be cleaner to just have this since that's the only thing we care about here because we overwrite `model_name` and `model_version`
```suggestion
if model_uri == "":
```

mingyu89: We get the three fields either all from jobs util or all from widgets. In case of manually running the notebook, we overwrite all three.
Sure. I don't see any difference and issues will be caught by assertions later.

arpitjasa-db (approved):
LGTM thanks @mingyu89 !

vladimirk-db (approved):

PR #58, opened 2023-02-22
[ML-27370] Add batch score inference using feature store

zhe-db:
Add support to use `score_batch` for batch inference if feature store enabled.

Tests: manual test with Azure-MLOps-CUJ


{{cookiecutter.project_name}}/steps/predict.py
```diff
@@ -1,7 +1,32 @@
...
+    """
+    Apply the model at the specified URI for batch inference on the table with name input_table_name,
+    writing results to the table with name output_table_name
+    """
+    table = spark_session.table(input_table_name)
 
+    prediction_df = fs_client.score_batch(
+        model_uri,
+        table
+    )
```
mingyu89, line 17: With or without feature store, the only difference is the predict step.
You can keep BatchInference file the same and only modify inside the predict_batch function.

```
...
table = spark_session.table(input_table_name)

{% if cookiecutter.include_feature_store == "yes" %}
    from databricks.feature_store import FeatureStoreClient

    fs_client = FeatureStoreClient()
    prediction_df = fs_client.score_batch(
        model_uri,
        table
    )
    prediction = prediction_df["prediction"]
{% else %}
    predict = mlflow.pyfunc.spark_udf(
        spark_session, model_uri, result_type="string", env_manager="conda"
    )
    prediction = predict(struct(*table.columns))
{% endif %}

    output_df = (
        table.withColumn("prediction", prediction)
    ...
```

zhe-db: sounds good

vladimirk-db: Aren't all the steps cleaned up if feature store is enabled? Can you please double check?

mingyu89: @vladimirk-db Great question! We are leaving the steps in the repo when feature store is enabled. Partial reason is that the folder also contains functions not related to recipes, for example, deploy.py, utils.py.
After the refactor-layout PR, the steps folder will only contain recipes functions and we can remove it.

vladimirk-db: Sounds good, just making sure we don't prune this file by accident :-)

vladimirk-db (approved):

{{cookiecutter.project_name}}/steps/predict.py
```diff
@@ -10,15 +10,29 @@ def predict_batch(
...
         table.withColumn("prediction", predict(struct(*table.columns)))
         .withColumn("model_version", lit(model_version))
         .withColumn("inference_timestamp", to_timestamp(lit(ts)))
     )
+    {% endif %}
     output_df.display()
     # Model predictions are written to the Delta table provided as input.
     # Delta is the default format in Databricks Runtime 8.0 and above.
-    output_df.write.format("delta").mode("overwrite").saveAsTable(output_table_name)
+    output_df.write.format("delta").mode("overwrite").saveAsTable(output_table_name)
```
arpitjasa-db, line 38: newline

{{cookiecutter.project_name}}/steps/predict.py
```diff
@@ -10,15 +10,29 @@ def predict_batch(
     writing results to the table with name output_table_name
     """
     table = spark_session.table(input_table_name)
-    predict = mlflow.pyfunc.spark_udf(
```
arpitjasa-db, line 13: Shouldn't this be a part of the `else`?

PR #59, opened 2023-02-24
Update validation run name to training run name plus suffix. Add training run link to validation run description

mingyu89:
This is the effort to better associate the validation run with the training run since they are in the same experiment.

## Design Doc
One decision doc https://docs.google.com/document/d/1KK7G3qz3tNc7ZPmCaZPRJ1fDVH5SyhFMtCJ8aWk4utM/edit#

### Run the training workflow

Screen Shot 2023-02-23 at 11 08 42 PM

### Validation run name = training run name + "-validation" 

Screen Shot 2023-02-23 at 11 08 55 PM

### The validation run correctly link to the training run

Screen Shot 2023-02-23 at 11 23 57 PM

### The corresponding example project commit 
https://github.com/databricks/mlops-azure-cuj/commit/7ee0c3f86ff5b6c389fa4e322a2d98fb899e6c47

### DIFF
--- a/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
+++ b/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
@@ -172,15 +172,38 @@ evaluator_config = {}
 
 # COMMAND ----------
 
-eval_result = None
-err = None
+# helper methods
+def get_run_link(run_info):
+    return "[Run](#mlflow/experiments/{0}/runs/{1})".format(
+        run_info.experiment_id, run_info.run_id
+    )
+
+
+def get_training_run(model_name, model_version):
+    model_versions = client.search_model_versions(f"name='{model_name}'")
+    training_run_id_optional = [
+        version.run_id for version in model_versions if version.version == model_version
+    ]
+    if not training_run_id_optional:
+        print("Unable to find the training run!")
+        return None
+    return mlflow.get_run(run_id=training_run_id_optional[0])
+
+
+def generate_run_name(training_run):
+    return None if not training_run else training_run.info.run_name + "-validation"
+
+
+def generate_description(training_run):
+    return (
+        None
+        if not training_run
+        else "Model Training Details: {0}\n".format(get_run_link(training_run.info))
+    )
 
 
 def log_to_model_description(run, success):
-    run_info = run.info
-    run_link = "[Run](#mlflow/experiments/{0}/runs/{1})".format(
-        run_info.experiment_id, run_info.run_id
-    )
+    run_link = get_run_link(run.info)
     description = client.get_model_version(model_name, model_version).description
     status = "SUCCESS" if success else "FAILURE"
     if description != "":
@@ -193,9 +216,14 @@ def log_to_model_description(run, success):
     )
 
 
-# run evaluate
-with mlflow.start_run() as run, tempfile.TemporaryDirectory() as tmp_dir:
+# COMMAND ----------
 
+training_run = get_training_run(model_name, model_version)
+# run evaluate
+with mlflow.start_run(
+    run_name=generate_run_name(training_run),
+    description=generate_description(training_run),
+) as run, tempfile.TemporaryDirectory() as tmp_dir:
     validation_thresholds_file = os.path.join(tmp_dir, "validation_thresholds.txt")
     with open(validation_thresholds_file, "w") as f:
         if validation_thresholds:
@@ -206,7 +234,6 @@ with mlflow.start_run() as run, tempfile.TemporaryDirectory() as tmp_dir:
                     )
                 )
     mlflow.log_artifact(validation_thresholds_file)
-
     try:
         eval_result = mlflow.evaluate(
             model=model_uri,
--- a/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
+++ b/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
@@ -180,14 +180,8 @@ def get_run_link(run_info):
 
 
 def get_training_run(model_name, model_version):
-    model_versions = client.search_model_versions(f"name='{model_name}'")
-    training_run_id_optional = [
-        version.run_id for version in model_versions if version.version == model_version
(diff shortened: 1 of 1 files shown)


{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
```diff
@@ -193,9 +216,14 @@ def log_to_model_description(run, success):
...
 
 
-# run evaluate
-with mlflow.start_run() as run, tempfile.TemporaryDirectory() as tmp_dir:
+# COMMAND ----------
 
+training_run = get_training_run(model_name, model_version)
+# run evaluate
+with mlflow.start_run(
+    run_name=generate_run_name(training_run),
```
mingyu89, line 224: If the training run is not found, both `run_name` and `description` will be None, which means random run name and empty description.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
```diff
@@ -172,15 +172,38 @@ def get_targets_from_recipe():
...
+        run_info.experiment_id, run_info.run_id
+    )
 
 
-def log_to_model_description(run, success):
-    run_info = run.info
-    run_link = "[Run](#mlflow/experiments/{0}/runs/{1})".format(
-        run_info.experiment_id, run_info.run_id
+def get_training_run(model_name, model_version):
+    model_versions = client.search_model_versions(f"name='{model_name}'")
```
vladimirk-db, line 183: Do we need to search model versions here? Why not look up the model version directly?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
```diff
@@ -172,15 +172,38 @@ def get_targets_from_recipe():
...
+        version.run_id for version in model_versions if version.version == model_version
+    ]
+    if not training_run_id_optional:
+        print("Unable to find the training run!")
+        return None
+    return mlflow.get_run(run_id=training_run_id_optional[0])
+
+
+def generate_run_name(training_run):
+    return None if not training_run else training_run.info.run_name + "-validation"
```
vladimirk-db, line 194: Maybe we shouldn't assume there only one validation run for a training run. What happens if users want to try out different metrics or different validation sets, but don't want to go through retraining? Or if the validation run fails with a retriable error (for whatever reason)?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
mingyu89: I investigated it for some time. This is the only way that I could find. 

The purpose is to get the training run info. We put its url in the validation run description and used its name as prefix.
However the mlflow recipe doesn't expose the training run information. We have to first get the registered model to find the run.
`Model.load("models:/prod-mlops-azure-cuj-model/83")` contains model metadata but doesn't have the run info.
```
databricks_runtime: 12.1.x-cpu-ml-scala2.12
flavors:
  python_function:
    artifacts:
      model_path:
        path: artifacts/sk_model
        uri: /root/.mlflow/recipes/77789c8c041dff48603f23e948b964c1e54512b9c0da38bf58dfba7bf19b9a1a/steps/train/outputs/sk_model
    cloudpickle_version: 2.0.0
    code: code
    env:
      conda: conda.yaml
      virtualenv: python_env.yaml
    loader_module: mlflow.pyfunc.model
    python_model: python_model.pkl
    python_version: 3.9.5
  sklearn:
    code: code
    pickled_model: artifacts/sk_model/model.pkl
    serialization_format: cloudpickle
    sklearn_version: 1.2.1
mlflow_version: 2.1.1
model_uuid: 2ab6371e9fa146c6bdb603e23633ce76
signature:
  inputs: '[{"name": "tpep_pickup_datetime", "type": "datetime"}, {"name": "tpep_dropoff_datetime",
    "type": "datetime"}, {"name": "trip_distance", "type": "double"}, {"name": "pickup_zip",
    "type": "integer"}, {"name": "dropoff_zip", "type": "integer"}]'
  outputs: '[{"type": "tensor", "tensor-spec": {"dtype": "float32", "shape": [-1]}}]'
utc_time_created: '2023-02-20 09:11:54.732645'
```

vladimirk-db: Right, what I mean is can you just look up the model version directly and avoid iterating over them?

https://mlflow.org/docs/latest/python_api/mlflow.client.html#mlflow.client.MlflowClient.get_model_version

mingyu89: Tried and it works!

mingyu89: There's exception if the version doesn't exist. Now I don't need to handle the error case.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
vladimirk-db: Discussed offline - this would work fine even in presence of multiple validation runs. LGTM

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
vladimirk-db: Yeah, but we are doing some exception handling anyways now (training run not found). It's not that big of a change.

There could be models with tons of model versions (thousands) - it's wasteful to fetch all and iterate on the client IMHO.

vladimirk-db (approved):

PR #60, opened 2023-02-28
Update pre_gen_project.py to use consistent format strings

DayneSorvisto:
Added replacement fields or use a normal string instead of an f-string.

Line 140 "f-string" doesn’t contain any replacement field, which probably means that some curly braces are missing and rest of file uses f-strings for consistency.


arpitjasa-db (commented):
This is great thanks @DayneSorvisto ! Would you mind manually generating this file to confirm the `ValueError` appears as expected?

DayneSorvisto:
> This is great thanks @DayneSorvisto ! Would you mind manually generating this file to confirm the `ValueError` appears as expected?

Happy to help on a great project.

I manually ran file to generate ValueError confirmed output appears as expected: https://gist.github.com/DayneSorvisto/ae88c7c47a34319297b9eae9c427c92f

arpitjasa-db (approved):
Perfect thank you! LGTM

PR #65, opened 2023-03-02
Model validation improvements

mingyu89:
I kept all the example code in model validation. Because the project itself is based on MLflow recipe example.

The PR contains the following changes:
- Extract all parameters from notebook `ModelValidation.py` to terraform file
- Create validation.py that defines methods that return custom metrics and validation thresholds.
- If fs is enabled : hide env parameter and show targets and model_type 
- If fs is disabled : show env parameter and hide targets and model_type 
- Update doc
- print message in the notebook about run_mode

There will be followup PR to handle the the parameter that can be directly inferred from MLflow recipe. 

---
### test 
Test 1 : Generate project without feature store, copy terraform files, validation.py and ModelValidation notebook to CUJ project and then run workflow. Test looks good.
https://adb-3253686410953403.3.azuredatabricks.net/?o=3253686410953403#job/741125644266775/run/1675013

~Test 2 : Generate project with feature store, copy terraform files, validation.py and ModelValidation notebook to CUJ project and then run workflow.~ The test needs some followup.

---

Make the same change in example repo https://github.com/databricks/mlops-azure-cuj/commit/dd092f62e4a054fcfc21151690456fd4df43d7f6
Manually run the notebook
Screen Shot 2023-03-01 at 7 35 32 PM

Run training workflow
Screen Shot 2023-03-01 at 8 27 56 PM
Screen Shot 2023-03-01 at 8 28 27 PM


{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
```diff
@@ -116,59 +131,40 @@ def get_targets_from_recipe():
...
-enable_baseline_comparison = False
+# take input
+enable_baseline_comparison = enable_baseline_comparison()
 
-# model validation data input for staging or prod workspace. A Pandas DataFrame or Spark DataFrame, containing evaluation features and labels.
-# Please refer to data parameter in mlflow.evaluate documentation https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.evaluate
 if env == "prod":
-    # TODO(required) : set data for prod workspace
-    data = None
+    data = get_prod_workspace_validation_input(spark)
```
mingyu89, line 148: I have to pass the `spark` object.
When I used `from databricks.sdk.runtime import spark` in the input python file, it works fine when I manually run the notebook but throws error when I run the workflow. 
It's probably related to MLR version.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/model_validation_input.py
```diff
@@ -0,0 +1,108 @@
...
+    ENABLED = 1
+    DRY_RUN = 2
+    DISABLED = 3
+
+
+# The `run_mode` defines whether model validation is enabled or not. It can be one of the three values:
+# - `DISABLED` : Do not run the model validation notebook.
+# - `DRY_RUN`  : Run the model validation notebook. Ignore failed model validation rules and proceed to move model to Production stage.
+# - `ENABLED`  : Run the model validation notebook. Move model to Production stage only if all model validation rules are passing.
+# TODO(required)
```
zhe-db, line 19: Nit: moving TODO before the documentation for better readability?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/terraform/prod/training-job.tf
```diff
@@ -51,12 +51,6 @@ resource "databricks_job" "model_training_job" {
       base_parameters = {
         env             = local.env
         experiment_name = databricks_mlflow_experiment.experiment.name
-        # Run mode for model validation. Possible values are :
-        #   disabled : Do not run the model validation notebook.
-        #   dry_run  : Run the model validation notebook. Ignore failed model validation rules and proceed to move model to Production stage.
-        #   enabled  : Run the model validation notebook. Move model to Production stage only if all model validation rules are passing.
-        # Please complete the TODO sessions in notebooks/ModelValidation before enabling model validation
-        run_mode = "disabled"
```
zhe-db, line 59: Why did we move `run_mode` out of terraform and hard code in the python code?

zhe-db (approved):

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/terraform/prod/training-job.tf
mingyu89: It's to keep all parameters in one place. But looks like the staging and prod workspace now share the same set of parameters.
I'm now thinking it may be better to have two input files or at least have two parameters for run_mode in staging and prod workspace.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/model_validation_input.py
```diff
@@ -0,0 +1,108 @@
...
+# Baseline model is a requirement for relative change and absolute change validation rules.
+# TODO(required)
+def enable_baseline_comparison():
+    return False
+
+
+# model validation data input for prod workspace. A Pandas DataFrame or Spark DataFrame, containing evaluation features and labels.
+# Please refer to data parameter in mlflow.evaluate documentation https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.evaluate
+# TODO(required)
+def get_prod_workspace_validation_input(spark):
```
vladimirk-db, line 35: Feels weird to have separate prod and staging validation parameters in here.

Why isn't this a get_workspace_validation_input - this may become an issue for the other parameters, too.

I guess the question is how to expose these arguments cleanly (considering some of the arguments are not literalls, they can be python code).

mingyu89: @vladimirk-db  Yeah I agree. 
There are several options for the API.
First question is which parameters will be shared by both staging and prod and which parameters are specific to one workspace. (If none of the parameters can be shared, we can just create two python input files)

My feeling is that the `model type`, `targets`, `custom metrics`, `thresholds` can be shared.
validation input `data` and `run mode` can not be shared.
`enable_baseline_comparison` maybe, maybe not.(always enable if there's a Production version?)

If only two parameters are different. we can either create separate methods(like I did).
Or pass Env as parameter, for example,
```
Class Env(Enum):
    DEV=1
    STAGING=2
    PROD=3

def get_workspace_validation_input(spark, env):
    if env == Env.PROD:
        return spark.sql(
            "SELECT * FROM delta.`dbfs:/databricks-datasets/nyctaxi-with-zipcodes/subsampled`"
        )
    elif env == Env.STAGING:
        return spark.sql(
            "SELECT * FROM delta.`dbfs:/databricks-datasets/nyctaxi-with-zipcodes/subsampled`"
        )
    return None

def get_workspace_run_mode(env)
    return RunMode.ENABLED if env == Env.PROD else RunMode.DISABLED
```

Or we can add the env parameter to all functions and let user decide. How you feel?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/model_validation_input.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+
+# model validation data input for staging workspace. A Pandas DataFrame or Spark DataFrame, containing evaluation features and labels.
+# Please refer to data parameter in mlflow.evaluate documentation https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.evaluate
+# TODO(required)
+def get_staging_workspace_validation_input(spark):
+    return spark.sql(
+        "SELECT * FROM delta.`dbfs:/databricks-datasets/nyctaxi-with-zipcodes/subsampled`"
+    )
```
vladimirk-db, line 47: Why isn't input a table name? No need to pass spark contexts if that's the case.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/model_validation_input.py
```diff
@@ -0,0 +1,108 @@
...
+# Please refer to targets parameter in mlflow.evaluate documentation https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.evaluate
+# TODO(required) : targets
+def get_targets():
+    return "mean_squared_error"
+
+
+# Custom metrics to be included. Set it to None if custom metrics are not needed.
+# Please refer to custom_metrics parameter in mlflow.evaluate documentation https://mlflow.org/docs/latest/python_api/mlflow.html#mlflow.evaluate
+# TODO(optional) : define custom metric function to be included in custom_metrics.
+def squared_diff_plus_one(eval_df, _builtin_metrics):
```
vladimirk-db, line 67: Why don't we fold this logic inside get_custom_metrics()? Then it becomes simpler in terms of understanding exactly what the inputs are to the notebook.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/model_validation_input.py
vladimirk-db: Sorry for the delay. Thinking about this, we should be consistent in how we handle user-provided code + configs across all the components in Stacks (training, validation, deployment, feature engineering, monitoring).

I'd rather not introduce env-based input handling just for the validation step.

Maybe we can write a quick one decision doc about this, what do you think @mingyu89  (would be useful for  Stacks overall going forward IMHO).

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
```diff
@@ -131,40 +153,43 @@ def get_targets_from_recipe():
...
+assert validation_thresholds_loader_function_name
+custom_metrics_loader_function = getattr(
+    importlib.import_module("validation"), custom_metrics_loader_function_name
+)
+validation_thresholds_loader_function = getattr(
+    importlib.import_module("validation"), validation_thresholds_loader_function_name
+)
+custom_metrics = custom_metrics_loader_function()
+validation_thresholds = validation_thresholds_loader_function()
+evaluator_config = {}
```
vladimirk-db, line 192: Will this get updated later? We're not using get_evaluator_config()

vladimirk-db (approved):
Looks good, thanks for making these changes!

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/validation/notebooks/ModelValidation.py
mingyu89: Yes! I'll update before merging. I shouldn't have left it out.

PR #66, opened 2023-03-06
Fix feature store notebook flow

arpitjasa-db:
As part of the single-workspace compatibility, we renamed the parent directory to include an environment suffix but did not update this for experiment references, especially in the feature store notebook. We also simplify some of the workflows and fix input/output pairs.

### DIFF
--- a/tests/test_create_project.py
+++ b/tests/test_create_project.py
@@ -112,7 +112,7 @@ def test_no_databricks_doc_strings_before_project_generation():
     assert_no_disallowed_strings_in_files(
         file_paths=test_paths,
         disallowed_strings=[
-            "https://docs.microsoft.com/en-us/azure/databricks",
+            "https://learn.microsoft.com/en-us/azure/databricks",
             "https://docs.databricks.com/",
             "https://docs.gcp.databricks.com/",
         ],
--- a/tests/test_create_project.py
+++ b/tests/test_create_project.py
@@ -220,8 +220,8 @@ def test_generate_succeeds_with_valid_params(tmpdir, valid_params):
 @pytest.mark.parametrize(
     "experiment_parent_dir,expected_dir",
     [
-        ("/mlops-project-directory/", "/mlops-project-directory-prod"),
-        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project-prod"),
+        ("/mlops-project-directory/", "/mlops-project-directory-${local.env}"),
+        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project-${local.env}"),
     ],
 )
 def test_strip_slash_if_needed_from_mlflow_experiment_parent_dir(
@@ -342,7 +342,7 @@ def test_generate_project_default_project_name_params(tmpdir):
         / "terraform/prod/locals.tf"
     ).read_text("utf-8")
     assert (
-        f'mlflow_experiment_parent_dir = "/{DEFAULT_PROJECT_NAME}-prod"'
+        f'mlflow_experiment_parent_dir = "/{DEFAULT_PROJECT_NAME}-${{local.env}}"'
         in tf_config_contents
     )
 
--- a/tests/test_create_project.py
+++ b/tests/test_create_project.py
@@ -223,7 +223,7 @@ def test_generate_succeeds_with_valid_params(tmpdir, valid_params):
         ("/mlops-project-directory/", "/mlops-project-directory-${local.env}"),
         (
             "/Users/test@databricks.com/project/",
-            "/Users/test@databricks.com/project-${local.env}"
+            "/Users/test@databricks.com/project-${local.env}",
         ),
     ],
 )
--- a/tests/test_create_project.py
+++ b/tests/test_create_project.py
@@ -221,7 +221,10 @@ def test_generate_succeeds_with_valid_params(tmpdir, valid_params):
     "experiment_parent_dir,expected_dir",
     [
         ("/mlops-project-directory/", "/mlops-project-directory-${local.env}"),
-        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project-${local.env}"),
+        (
+                "/Users/test@databricks.com/project/",
+                "/Users/test@databricks.com/project-${local.env}"
+        ),
     ],
 )
 def test_strip_slash_if_needed_from_mlflow_experiment_parent_dir(
--- a/tests/test_create_project.py
+++ b/tests/test_create_project.py
@@ -222,8 +222,8 @@ def test_generate_succeeds_with_valid_params(tmpdir, valid_params):
     [
         ("/mlops-project-directory/", "/mlops-project-directory-${local.env}"),
         (
-                "/Users/test@databricks.com/project/",
-                "/Users/test@databricks.com/project-${local.env}"
+            "/Users/test@databricks.com/project/",
+            "/Users/test@databricks.com/project-${local.env}"
         ),
     ],
 )
--- a/tests/test_create_project.py
+++ b/tests/test_create_project.py
@@ -220,8 +220,8 @@ def test_generate_succeeds_with_valid_params(tmpdir, valid_params):
 @pytest.mark.parametrize(
     "experiment_parent_dir,expected_dir",
     [
-        ("/mlops-project-directory/", "/mlops-project-directory"),
-        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project"),
+        ("/mlops-project-directory/", "/mlops-project-directory-prod"),
+        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project-prod"),
     ],
 )
 def test_strip_slash_if_needed_from_mlflow_experiment_parent_dir(
@@ -342,7 +342,7 @@ def test_generate_project_default_project_name_params(tmpdir):
         / "terraform/prod/locals.tf"
     ).read_text("utf-8")
     assert (
-        f'mlflow_experiment_parent_dir = "/{DEFAULT_PROJECT_NAME}"'
+        f'mlflow_experiment_parent_dir = "/{DEFAULT_PROJECT_NAME}-prod"'
         in tf_config_contents
     )
 
--- a/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
+++ b/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
@@ -24,9 +24,9 @@ jobs:
           python-version: 3.8
       # Feature store tests bring up a local Spark session, so Java is required.
       - uses: actions/setup-java@v2
-      with:
-        distribution: 'temurin'
-        java-version: '11'
+        with:
+          distribution: 'temurin'
+          java-version: '11'
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
--- a/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
+++ b/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
@@ -122,9 +122,9 @@ jobs:
                 "notebook_task": {
                   "notebook_path": "{{cookiecutter.project_name}}/feature_engineering/notebooks/TrainWithFeatureStore",
                   "base_parameters": {
-                    "test_mode": "True",
+                    "env": "staging"
                     "training_data_path": "/databricks-datasets/nyctaxi-with-zipcodes/subsampled",
-                    "experiment_name": "{{cookiecutter.mlflow_experiment_parent_dir}}/{{cookiecutter.experiment_base_name}}-test",
+                    "experiment_name": "{{cookiecutter.mlflow_experiment_parent_dir}}-staging/{{cookiecutter.experiment_base_name}}-test",
                     "model_name": "{{cookiecutter.model_name}}-test"
                   }
                 },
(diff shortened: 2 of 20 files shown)


{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
```diff
@@ -120,11 +120,11 @@ jobs:
...
                 ],
                 "notebook_task": {
-                  "notebook_path": "{{cookiecutter.project_name}}/feature_engineering/notebooks/TrainWithFeatureStore",
+                  "notebook_path": "{{cookiecutter.project_name}}/training/notebooks/TrainWithFeatureStore",
                   "base_parameters": {
-                    "test_mode": "True",
+                    "env": "staging"
                     "training_data_path": "/databricks-datasets/nyctaxi-with-zipcodes/subsampled",
-                    "experiment_name": "{{cookiecutter.mlflow_experiment_parent_dir}}/{{cookiecutter.experiment_base_name}}-test",
+                    "experiment_name": "{{cookiecutter.mlflow_experiment_parent_dir}}-staging/{{cookiecutter.experiment_base_name}}-test",
```
mingyu89, line 127: Aren't we supposed to use dev workspace for testing(CI)?

mingyu89 (commented):
The change looks good to me. Thanks for doing the testing for feature store.
(I didn't test feature store in real workspace during project layout change)

Just one quick question in the code.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
arpitjasa-db: The `DATABRICKS_HOST` value specified in the file is `staging` so I just used that. We don't technically take in a dev workspace parameter in stacks so setting up a dev workflow from default would involve that

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/training/notebooks/TrainWithFeatureStore.py
```diff
@@ -30,11 +34,11 @@
 # MLflow experiment name.
 dbutils.widgets.text(
     "experiment_name",
-    "/{{cookiecutter.project_name}}/{{cookiecutter.experiment_base_name}}-test",
+    f"/{{cookiecutter.mlflow_experiment_parent_dir}}-{env}/{{cookiecutter.experiment_base_name}}-test",
```
mingyu89, line 37: Actually, the `mlflow_experiment_parent_dir` is input from user. 
Now we are adding suffix to the parent dir name. It may be a bit unexpected. Can we move the `-{env}` to the experiment name instead of experiment folder?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
mingyu89: If in test model(not initiated from terraform workflow), we may want to just use "/{{cookiecutter.experiment_base_name}}-test".
Because if user run the notebook from dev workspace, the folder  `{{cookiecutter.mlflow_experiment_parent_dir}}` or `{{cookiecutter.mlflow_experiment_parent_dir}}-staging` will not exist. The experiment name `{cookiecutter.mlflow_experiment_parent_dir}}-staging/<name>` may cause error.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/training/profiles/databricks-dev.yaml
```diff
@@ -1,12 +1,12 @@
 # This profile contains config overrides for model development on Databricks
 experiment:
-  name: "/{{cookiecutter.experiment_base_name}}-dev"
+  name: "/dev-{{cookiecutter.experiment_base_name}}"
```
mingyu89, line 3: I think the reason it doesn't have the parent dir is because in dev workspace, the parent dir doesn't exist. This is the same for running the notebook in test model.

mingyu89 (changes requested):
After thinking twice, I added a concern. Thanks.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/training/notebooks/TrainWithFeatureStore.py
arpitjasa-db: No because we generate the `mlflow_experiment_parent_dir` with a `-staging` suffix in the staging env, and a `-prod` suffix in the prod env to allow for single-workspace mode operation. So there is no directory with just `mlflow_experiment_parent_dir`

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/training/profiles/databricks-dev.yaml
arpitjasa-db: Yeah I'm just moving the suffix to prefix to match naming conventions for `staging` and `prod`

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/training/notebooks/TrainWithFeatureStore.py
mingyu89: @arpitjasa-db We may not need to update the parent dir with the {env}, since the experiment name is already different.
The mlflow-experiment.tf
```
resource "databricks_mlflow_experiment" "experiment" {
  name        = "${local.mlflow_experiment_parent_dir}/${local.env_prefix}{{cookiecutter.experiment_base_name}}"
  description = "MLflow Experiment used to track runs for {{cookiecutter.project_name}} project."
}
```
For example, if parent dir input from user is `parent-dir` and project_name is `project`. Then after the change you could have the following if all three points to the same workspace
 - /parent-dir-prod/prod-project : from terraform prod
 - /parent-dir-prod/project-test : manually run notebook and select prod
 - /parent-dir-staging/staging-project : from terraform staging or github test
 - /parent-dir-staging/project-test : manually run notebook and select prod
 - /dev-project : manually run notebook and select dev

I may have missed what the PR is trying to resolved. How about keeping the parent dir the same as user input?
- /parent-dir-prod/prod-project : from terraform prod
- /parent-dir-prod/staging-project  : from terraform staging
- /dev-project : manually run notebook, no matter what env is selected
- /test-project : github test

mingyu89: If user manually run notebook `TrainWithFeatureStore.py` and select dev as env, I think it will throw error saying directory not found because dir `/parent-dir-dev/` doesn't exist.
As the list mentioned above, `/parent-dir-dev/project-test` is not there because `parent-dir-dev` will not be created by terraform.

arpitjasa-db: @mingyu89 in your second example you still call the directory `parent-dir-prod` but assuming you meant to keep it just `parent-dir`, the problem is in Terraform we create this directory via the modules. And in single-workspace mode we were creating two TF directory resources that pointed to the same directory which was overwriting each other's permissions and other resources. It's cleaner to just separate them. This was already done in a previous PR. 

In this PR, our flow is currently broken as I've seen during dogfooding, so I'm fixing it by passing in the env to the directory name, that's all this PR is for.

arpitjasa-db: > If user manually run notebook `TrainWithFeatureStore.py` and select dev as env, I think it will throw error saying directory not found because dir `/parent-dir-dev/` doesn't exist. As the list mentioned above, `/parent-dir-dev/project-test` is not there because `parent-dir-dev` will not be created by terraform.

Also for this yes, but we don't create a directory in the `dev` env anyways since we don't take in a dev workspace/create TF resources, so either way they'd need to manually create that directory in their dev env.

mingyu89: @arpitjasa-db Ah I see it's so tricky. And in my example I meant `parent-dir`, updated.
If the parent dir must be different, how about keep the experiment so that we don't have `staging` and `prod` in two different places?

- /parent-dir-prod/project : from terraform prod
- /parent-dir-staging/project : from terraform staging
- /dev-project : manually run notebook, no matter what env is selected
- /test-project : github test

arpitjasa-db: Yeah the only issue with that would be in single-workspace mode, in the experiment observatory (or model registry if we do the same for models), the two experiments/models would look identical

mingyu89: You are right. We only show the name on UI. Then my only `nit` is to but the "prod" on the same side, 
`/parent-dir-prod/project-prod` instead of `/parent-dir-prod/prod-project`.

mingyu89 (approved):

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/training/notebooks/TrainWithFeatureStore.py
mingyu89: And we are using the staging workspace for `run-tests-fs.yml`, the parent dir makes sense.
Can we remove `dev` from this notebook? It has no use and only causes confusion.
After that we'll have
 - /parent-dir-prod/project-prod : from terraform prod
 - /parent-dir-prod/project-test : manually run notebook with "prod"
 - /parent-dir-staging/project-staging : from terraform staging
 - /parent-dir-staging/project-test : manually run notebook with "staging" or CI

arpitjasa-db: Yeah sure makes sense to me, I'll remove it and set default to `staging`

vladimirk-db (approved):

arpitjasa-db:
Hey @vladimirk-db and @mingyu89 I updated some more of the feature store inconsistencies in this PR, and would like yall to take another look before I merge to make sure that I don't just blast forward with these changes haha

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/docs/ml-pull-request.md
```diff
@@ -34,7 +34,7 @@ is planned for the future.
 ## Viewing test status and debug logs
 Opening a pull request will trigger a 
 {%- if cookiecutter.cicd_platform == "gitHub" -%} 
-[workflow](../.github/workflows/run-tests.yml) 
+[workflow](../.github/workflows/run-tests{% if cookiecutter.include_feature_store %}-fs{% endif %}.yml)
```
mingyu89, line 37:
```
if cookiecutter.include_feature_store == "yes"
```

tests/test_create_project.py
```diff
@@ -220,8 +220,8 @@ def test_generate_succeeds_with_valid_params(tmpdir, valid_params):
 @pytest.mark.parametrize(
     "experiment_parent_dir,expected_dir",
     [
-        ("/mlops-project-directory/", "/mlops-project-directory"),
-        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project"),
+        ("/mlops-project-directory/", "/mlops-project-directory-prod"),
+        ("/Users/test@databricks.com/project/", "/Users/test@databricks.com/project-prod"),
```
mingyu89, line 224: The new expected parent dirs are 
```
/mlops-project-directory-${local.env}
/Users/test@databricks.com/project-${local.env}
```

tests/test_create_project.py
```diff
@@ -342,7 +342,7 @@ def test_generate_project_default_project_name_params(tmpdir):
         / "terraform/prod/locals.tf"
     ).read_text("utf-8")
     assert (
-        f'mlflow_experiment_parent_dir = "/{DEFAULT_PROJECT_NAME}"'
+        f'mlflow_experiment_parent_dir = "/{DEFAULT_PROJECT_NAME}-prod"'
```
mingyu89, line 345: `/{DEFAULT_PROJECT_NAME}-${local.env}`

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/.github/workflows/run-tests-fs.yml
```diff
@@ -22,6 +22,11 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.8
+      # Feature store tests bring up a local Spark session, so Java is required.
+      - uses: actions/setup-java@v2
+      with:
```
mingyu89, line 27: The indentation is causing test failure.

mingyu89 (approved):
LGTM

vladimirk-db (approved):
Thanks for the fix, it looks cleaner as well now.

PR #68, opened 2023-03-15
Fix model deployment

mingyu89:
The model deployment was broken due to model layout change. Found by 
@niall-turbitt https://docs.google.com/document/d/1ChQrfnRxN2CAhECX7xJZw_tmu8PoweI9AlLqgx67Tmk/edit#

The change is to update github action to correctly call deploy and fix the import of utils.

---

I'm not importing "{{cookiecutter.project_name}}.utils" because currently there's no guarantee that the project_name is also a valid package name.(For example, the "mlops-azure-cuj" is not a valid package name)

After I added the `__init__` files, the relative import of "../.." works on my laptop but doesn't work for github actions(maybe because of the current work directory).

I have to use absolute path to import utils and it's working. 

---

Tested on CUJ project. The CUJ deployment was also broken after the layout change yesterday and I didn't notice it. Now it's fixed. https://github.com/databricks/mlops-azure-cuj/actions/runs/4431460872/jobs/7774443532

This is the change made in CUJ project https://github.com/databricks/mlops-azure-cuj/pull/74/files/f0f4df8dec32780fc2a6bd5f317e637a13eac865..1034207c3c9547c825a8bfeaa4e717ad9da39d27

### DIFF
--- a/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/deployment/model_deployment/deploy.py
+++ b/{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/deployment/model_deployment/deploy.py
@@ -1,6 +1,9 @@
+import sys
+import pathlib
+
+sys.path.append(str(pathlib.Path(__file__).parent.parent.parent.resolve()))
 from utils import get_deployed_model_stage_for_env
 from mlflow.tracking import MlflowClient
-import sys
 
 
 def deploy(model_uri, env):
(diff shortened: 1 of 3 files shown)


{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name}}/deployment/model_deployment/deploy.py
```diff
@@ -1,6 +1,9 @@
+import sys
+import pathlib
+
+sys.path.append(str(pathlib.Path(__file__).parent.parent.parent.resolve()))
```
arpitjasa-db, line 4: The reason we added and moved the imports above this line is because Python wasn't able to discover and import from `utils`?

mingyu89: Yeah. I tried several ways. 
For CUJ project, the "current directory" will be the outer "mlops-azure-cuj" folder. The best way to import utils will be `mlops-azure-cuj.utils`. However `mlops-azure-cuj` is not a valid package name and it doesn't have to be. 
So I used absolute path to add the location of utils to `sys.path` and directly imported it.

vladimirk-db (approved):

arpitjasa-db (approved):
Sounds good, thanks for the quick fix @mingyu89 !

PR #77, opened 2023-05-19
Update MLOps Stacks to Use Databricks Asset Bundles

arpitjasa-db:


mingyu89 (approved):
🚀

vladimirk-db:
Do we want to tag the pre-bundles version with a release tag or something? In case users want to refer to it and use it for some reason.

arpitjasa-db:
> Do we want to tag the pre-bundles version with a release tag or something? In case users want to refer to it and use it for some reason.

Yeah I was thinking about that but they can also just refer to the commit right before this release. But also release tags are free so I can just do it too 😛

PR #82, opened 2023-05-31
Azure DevOps bundle cli installer update

niall-turbitt:
- Update Azure Pipelines to install Databricks CLI using `curl -fsSL https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh | sh` rather than the installer script within the repo
- Remove installer script
- Fix typo error on SP prod tenant ID


arpitjasa-db (approved):
LGTM just one question, I notice we don't supply architecture anymore, is that taken care of by the `install.sh` script?

niall-turbitt:
> LGTM just one question, I notice we don't supply architecture anymore, is that taken care of by the `install.sh` script?

@arpitjasa-db I assume so as I did not need to pass that as an env var when testing. @pietern can you confirm this?

vladimirk-db:
> > LGTM just one question, I notice we don't supply architecture anymore, is that taken care of by the `install.sh` script?
> 
> @arpitjasa-db I assume so as I did not need to pass that as an env var when testing. @pietern can you confirm this?

It's taken care of by the script:
https://raw.githubusercontent.com/databricks/setup-cli/main/install.sh

vladimirk-db (approved):
Thank you for the quick fix; LGTM, ship it!

pietern (commented):
@niall-turbitt Before merging, could you please update `[...]-tests-ci.yml` to run just `databricks version` instead of `databricks version --detail`? The latter is going away in favor of `--output json` to be consistent with other commands.

I believe this is the only callsite with the `--detail` flag.

niall-turbitt:
> @niall-turbitt Before merging, could you please update `[...]-tests-ci.yml` to run just `databricks version` instead of `databricks version --detail`? The latter is going away in favor of `--output json` to be consistent with other commands.
> 
> I believe this is the only callsite with the `--detail` flag.

done! And also removed the pre-existing env var declaring the version. Thanks for reviewing @pietern

PR #93, opened 2023-07-20
Replace MLflow Recipes as default option with Delta Tables

arpitjasa-db:
We update MLOps Stacks to use Delta Tables by default instead of MLflow Recipes.

Tested using generated stack here: https://github.com/databricks/test-delta-stacks, and confirmed conflicting options cannot be selected:
Screenshot 2023-07-28 at 11 06 51 PM


README.md
```diff
@@ -53,6 +53,10 @@ This will prompt for parameters for project initialization. Some of these parame
    Otherwise, user can leave it blank to use the name of the current project as root directory name.
  * ``cloud``: Cloud provider you use with Databricks (AWS, Azure, or GCP)
  * ``cicd_platform`` : CI/CD platform of choice (GitHub Actions or GitHub Actions for GitHub Enterprise Servers or Azure DevOps)
+ * ``framework``: Choose between Delta Tables (default), Feature Store, and MLflow Recipes.
+   - If ``Delta Tables`` is selected, will provide the default stack using Delta Tables for data input and output.
```
mingyu89, line 57: We've talked a lot offline. Let's bring the discussion online.

There are two parts of the problem:
1. Given the new option of non-feature-store non-recipe training, what will be the user facing questions to ask users? 2 yes/no questions or 1 enum selection?
2. For internal implementation, do we want to use 2 variable or one enum? 

### For question 2, 
I'm convinced that it will be cleaner in code if we use 1 enum to switch between feature-store / recipe / no-fs-no-recipe.

### For question 1, 
the tradeoffs are:

- If use two questions to ask if recipes / fs is enabled, 
    - recipe(yes) + fs(yes) is not a valid combination. 
    - one more question to ask users
    - (Maybe) easier to deprecate recipe from mlops stacks in the future?
- If use a single question
    - We have to find a better name of  `Delta Tables` because essentially all three option use delta table and input and output
        - any name suggestion? simple training notebook? 
    - fs and recipes are not the same level of option. fs uses different batch inference code, there's no model validation support for fs. mlflow recipe is only for training notebook.

cc @vladimirk-db  @dbczumar @sebastiaanboer-db @zhe-db

arpitjasa-db: Thanks for the write-up @mingyu89 !

I think if we can find a better name for `Delta Tables` and provide an accurate description, it would be the simplest to reason about with a single question. Asking two questions and then throwing on an invalid combo is not a great user experience. Plus it will inevitably lead to questions like "When will FS work with MLflow recipes?" 😂

vladimirk-db: Prefer option 1 for question 1.
use mlflow recipes [no is default]
use feature store [no is default]

If we can do conditional cookiecutter where if a user selects mlflow recipes, then the feature store question is omitted, it would be ideal. But it's not strictly required.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
```diff
@@ -1,140 +1,137 @@
...
+dbutils.widgets.dropdown("env", "staging", ["staging", "prod"], "Environment Name")
+env = dbutils.widgets.get("env")
 
-# COMMAND ----------
+# Path to the Hive-registered Delta table containing the training data.
+dbutils.widgets.text(
+    "training_data_path",
+    "/databricks-datasets/nyctaxi-with-zipcodes/subsampled",
+    label="Path to the training data",
+)
```
vladimirk-db, line 33: Why not use (fully qualified) table name here? It would be more UC-friendly as well, right?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
```diff
@@ -1,140 +1,137 @@
...
 
-from mlflow.recipes.utils import (
-    get_recipe_config,
-    get_recipe_root_path,
+# MLflow experiment name.
+dbutils.widgets.text(
+    "experiment_name",
+    f"/dev-{{cookiecutter.experiment_base_name}}",
```
vladimirk-db, line 38: Do we really want to hardcode the `dev-` prefix here?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
```diff
@@ -1,140 +1,137 @@
...
+    model, artifact_path="lgb_model", registered_model_name=model_name,
+)
 
-# COMMAND ----------
+# Build out the MLflow model registry URL for this model version.
+workspace_url = spark.conf.get("spark.databricks.workspaceUrl")
+model_version = get_latest_model_version(model_name)
+model_registry_url = "https://{workspace_url}/#mlflow/models/{model_name}/versions/{model_version}".format(
+    workspace_url=workspace_url, model_name=model_name, model_version=model_version
+)
```
vladimirk-db, line 130: where is this used?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
mingyu89: @vladimirk-db To make sure the example work out of the box, we need to take existing data as input from database. This `/databricks-datasets/nyctaxi-with-zipcodes/subsampled` file exists in dbfs.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
mingyu89: This is the default value when users manually run the notebook. We want to make it clear that it's not for CI/CD/staging/prod purpose. That's why we added the `dev-`
If it's from CI/CD, the experiment name parameter is passed to the notebook and will overwrite this default value.

README.md
zhe-db: Since recipe(yes) + fs(yes) is not a valid combination, do we want to skip the second question for fs if user give yes to recipes?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
vladimirk-db: (These are registered in UC actually: https://docs.databricks.com/dbfs/databricks-datasets.html#unity-catalog-datasets )

mingyu89: I didn't know that before! 
If we want to avoid adding a question about whether UC is enabled, we can defer the change to a point when we can detect whether UC is enabled during project initialization.

mingyu89: That's a feature like "Fully use UC in stacks", including input/output tables, models, monitoring, etc.

vladimirk-db: Yeah, let's defer this until we add UC-based asset templates

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
arpitjasa-db: I don't believe it is, but it's the same in the `TrainWithFeatureStore.py` notebook that this is modeled on. I believe we create this URL here so the user is aware of where the version is/what the URL looks like.

README.md
arpitjasa-db: Updated the code, conditional variable does not seem to be possible at this time.

mingyu89: Yeah

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
vladimirk-db: Ah, we may have returned it as a GH PR comment in the past.  If it's not used though, we could clean it up.

README.md
```diff
@@ -71,6 +67,8 @@ ready to productionize a model. We recommend specifying any known parameters upf
...
     repo pull ML code from this branch.
  * ``read_user_group``: User group name to give READ permissions to for project resources (ML jobs, integration test job runs, and machine learning resources). A group with this name must exist in both the staging and prod workspaces. Defaults to "users", which grants read permission to all users in the staging/prod workspaces. You can specify a custom group name e.g. to restrict read permissions to members of the team working on the current ML project.
+ * ``include_feature_store``: If selected, will provide [Databricks Feature Store](https://docs.databricks.com/machine-learning/feature-store/index.html) stack components including: project structure and sample feature Python modules, feature engineering notebooks, ML resource configs to provision and manage Feature Store jobs, and automated integration tests covering feature engineering and training.
+ * ``include_mlflow_recipes``: If selected, will provide [MLflow Recipes](https://mlflow.org/docs/latest/recipes.html) stack components, dividing the process into configurable steps and profiles.
```
vladimirk-db, line 71:
```suggestion
 * ``include_mlflow_recipes``: If selected, will provide [MLflow Recipes](https://mlflow.org/docs/latest/recipes.html) stack components, dividing the training pipeline into configurable steps and profiles.
```

vladimirk-db: `process` is too generic?

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/README.md
```diff
@@ -72,10 +72,10 @@ Otherwise, e.g. if iterating on ML code for a new project, follow the steps belo
   `dbx sync repo --profile {{cookiecutter.project_name}}-dev --source . --dest-repo your-repo-name`, where `your-repo-name` should be the last segment of the full repo name (`/Repos/username/your-repo-name`)
 
 #### Running code on Databricks
```
vladimirk-db, line 74: This whole file is talking about mlflow recipes, which is not the default option any more. I wonder if we should just provide this file when recipes is selected, and have a simpler README otherwise.

mingyu89:
It looks good to me. That's a lot of condition changes.

As one more step of sanity check, could you generate the mlflow recipes project and fs project using both master and your branch? Then do a diff of the two projects? Also do a diff of the training notebooks since they are renamed now.

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/notebooks/Train.py
arpitjasa-db: Cleaned it up!

{{cookiecutter.root_dir__update_if_you_intend_to_use_monorepo}}/{{cookiecutter.project_name_alphanumeric_underscore}}/training/README.md
arpitjasa-db: We only return this file when the recipes option is selected and otherwise we don't have a README since it's only a single notebook for the default and the feature store options. This is the same behavior we had for the feature store option prior to this change

vladimirk-db (approved):
LGTM, assuming @mingyu89's  testing comments are addressed.

Good stuff!

arpitjasa-db:
> It looks good to me. That's a lot of condition changes.
> 
> As one more step of sanity check, could you generate the mlflow recipes project and fs project using both master and your branch? Then do a diff of the two projects? Also do a diff of the training notebooks since they are renamed now.

Here's Feature Store: https://github.com/databricks/test-delta-stacks/pull/4 and MLflow Recipes: https://github.com/databricks/test-delta-stacks/pull/5

mingyu89 (approved):

mingyu89 (approved):
Looks good!

mabreuortega:
I'm having issues with this upgrade, cookiecutter no longer works and the suggested databricks cli version is not available

mingyu89:
@mabreuortega The databricks cli version is available here https://github.com/databricks/cli/releases/tag/v0.204.0

mingyu89:
@mabreuortega Created PR to explicitly add the link to databricks cli installation doc and releases page
https://github.com/databricks/mlops-stack/pull/100

PR #97, opened 2023-08-21
Fix all mlops stacks tests

mingyu89:
Fix unit tests.
Created the separate PR for easier review.

## Design doc and validation of migrating mlops stacks to use asset templates 
https://docs.google.com/document/d/19MOK9o-f0A_-NMmxC1cJGQSQ2WkJXZ4vXQZ4plu-tng/edit#bookmark=id.tsk81t8hk8bf


arpitjasa-db (approved):
Reviewed offline with @mingyu89 but confirmed that we are not losing test coverage and that we have fixed all the tests based on the passing jobs

PR #117, opened 2023-11-08
In the TrainWithFeatureStore notebook, set databricks-uc only when us model registry is used

mingyu89:


zhe-db (approved):

arpitjasa-db (approved):
Thanks for catching this

PR #121, opened 2023-11-30
Update template to work with CLI v0.210.1

pietern:
* Fix the issue where starting v0.210.0 the `permissions` block is a valid top-level field in the bundle configuration. This template uses the same field as an anchor with a schema incompatible with the bundle configuration schema. This mismatch means none of the bundle commands work with this template starting with v0.210.1. To fix this, I renamed the field with the anchor block to `common_permissions`.
* Pin the Databricks CLI to a fixed version such that once automation is set up, it will continue to work.


pietern:
Note: the whitespace changes are courtesy of VS Code settings:
```
  "files.insertFinalNewline": true,
  "files.trimFinalNewlines": true,
  "files.trimTrailingWhitespace": true,
```

andrewnester (approved):

arpitjasa-db (approved):
Thanks for doing this @pietern! Would you mind updating the docs where it says `0.209.1` to this new version as well?

pietern:
@arpitjasa-db Done; please TAL

arpitjasa-db (approved):
That's perfect thank you! I wonder if there's a way we can specify one location with the version and propagate it to make it easier in the future though

PR #126, opened 2024-01-03
Update README.md

pietern:
Removes duplicate line.


arpitjasa-db (approved):
Thanks for catching this!

Issue #124, opened 2023-12-12
`bundle init` asks for schema to use when registering a model in Unity Catalog when selected to not use Unity Catalog for Model Registry


woropajj:
`databricks bundle init mlops-stacks` prompts to set schema and privileges in Unity Catalog when I selected that I don't want to use Unity Catalog:

```bash
# other steps

Whether to use the Model Registry with Unity Catalog: no

Name of schema to use when registering a model in Unity Catalog. 
Note that this schema must already exist, and we recommend keeping the name the same as the project name as well as giving the service principals the right access. Default [my-mlops-project]: 

User group name to give EXECUTE privileges to models in Unity Catalog. A group with this name must exist in the Unity Catalog that the staging and prod workspaces can access. Default [account users]: 

Whether to include Feature Store: no

Whether to include MLflow Recipes: no

✨ Your MLOps Stack has been created in the 'my-mlops-project-no-unity' directory!

Please refer to the README.md of your project for further instructions on getting started.
```

Side question: will this stack work if I don't have Unity Catalog enabled at all in my workspace?

PR #132, opened 2024-01-09
Add Support for Conditional Initialization

arpitjasa-db:
Add support for conditional initialization for UC prompts as well as Recipes. Also update minimum CLI version.

Fixes https://github.com/databricks/mlops-stacks/issues/124

### DIFF
--- a/databricks_template_schema.json
+++ b/databricks_template_schema.json
@@ -1,12 +1,12 @@
 {
   "welcome_message": "Welcome to MLOps Stacks. For detailed information on project generation, see the README at https://github.com/databricks/mlops-stacks/blob/main/README.md.",
-  "min_databricks_cli_version": "v0.210.1",
+  "min_databricks_cli_version": "v0.211.0",
   "properties": {
     "input_project_name": {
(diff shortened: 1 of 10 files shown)


databricks_template_schema.json
```diff
@@ -98,7 +112,14 @@
       "type": "string",
       "description": "\nWhether to include MLflow Recipes",
       "default": "no",
-      "enum": ["no", "yes"]
+      "enum": ["no", "yes"],
+      "skip_prompt_if": {
+        "properties": {
```
arpitjasa-db, line 117: Note specifying multiple properties is an "AND" operation, not an "OR" operation. So we can't specify that Recipes should be skipped if either Unity Catalog or Feature Store is "yes". @shreyas-goenka can we add that support?

mingyu89 (approved):

databricks_template_schema.json
shreyas-goenka: Filed a ticket in our internal backlog to add support for OR

Issue #122, opened 2023-12-01
monorepo doesn't work as expected


lbcommer:
It's not possible to create more than one project in the same repository. 

Reproducibility steps:

1) Create a project `project-1` with `databricks bundle init mlops-stacks`
For the `input_root_dir` variable, I chose the default value of `my-mlops-project`. It worked ok.

2) Create a second project, `project-2`, keeping the same value for `input_root_dir`  (`my-mlops-project`)
That shows an error because it's trying to write on already existing files:

> Error: failed to initialize template, one or more files already exist: my-mlops-project\.gitignore

I am using the last available version of `databricks-cli` (0.210.1)

PR #133, opened 2024-01-09
Fix Monorepo Setup & Separate by Role

arpitjasa-db:
## Changes
This PR makes multiple changes in line with the [Separate by Role Project Proposal](https://docs.google.com/document/d/1eUE8UmcLh8vzWnH95mVB-QawoeOZ16L4WVrGDkk3qvs/edit?usp=sharing) to separate CICD generation from project generation to allow monorepos to work and making the division more intuitive:
- Add a new parameter `input_setup_cicd_and_project` that will allow three parameters: `CICD_and_Project` (the current default flow), `Project_Only` which sets up only project files, and `CICD_Only` which only sets up the root directory and provides a workflow to generate CI/CD files for each project in the directory.
- Update CI/CD workflows pull requests to only trigger if files within respective project folder are modified in a PR.
- Update docs to clarify how to set this up.

Fixes https://github.com/databricks/mlops-stacks/issues/122

Also includes various quality-of-life improvements and minor bug fixes such as constant config versions, clean-up of unused variables, and proper file generation.

### Project Generation Workflow:
Screenshot 2024-01-25 at 2 53 22 PM
Screenshot 2024-01-26 at 2 55 57 AM
Screenshot 2024-01-25 at 2 54 16 PM

### CICD Zip Generation Workflow
Screenshot 2024-02-03 at 2 08 59 AM


template/{{.input_root_dir}}/.azure/devops-pipelines/{{.input_project_name}}-bundle-cicd.yml.tmpl
```diff
@@ -10,8 +10,12 @@ trigger:
     include:
       - {{template `default_branch` .}}
       - {{template `release_branch` .}}
+  paths:
+    include:
+      - {{template `project_name_alphanumeric_underscore` .}}/*
 
 variables:
+  - workingDirectory: {{template `project_name_alphanumeric_underscore` .}}
```
arpitjasa-db, line 18: @niall-turbitt to test ADO changes. Would you mind also generating and testing a monorepo setup with ADO to confirm it works/triggers as expected?

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
```diff
@@ -1,5 +1,144 @@
 # {{template `project_name` .}}
 
+{{- if (eq .input_setup_cicd `no`) }}
+This directory contains an ML project based on the default
+[Databricks MLOps Stacks](https://github.com/databricks/mlops-stacks),
+defining a production-grade ML pipeline for automated retraining and batch inference of an ML model on tabular data.
+
+
+## Code structure
+This project contains the following components:
```
mingyu89, line 10: This readme file is only relevant when users have done the step1 of "Setting up CI/CD for multiple projects" and before doing step2.
And its content is not consistent with project-overview.md. 
Let's avoid mentioning the intermediate state dir structure.

mingyu89 (dismissed):
The approach of copying the CICD files of the first project will not work well because it assumes that all the projects have the same CICD platform and feature store selections, which may not be the case.

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
arpitjasa-db: The idea is that they won't have the `docs` folder for this project (since it's a top-level directory) which is fine since most of them deal with CI/CD. I copied the non-CI/CD relevant stuff to this project so I think the dir structure helps since it is project specific in this case.

And that is correct this file is relevant only after Step 1, but step 1 is generating the project so by definition they won't have this file until they do step 1. And step 2 is linking the CI/CD workflows, which is also the last step. If they do it without the help of this README awesome! And if not, well then that's what the README is for!

> The approach of copying the CICD files of the first project will not work well because it assumes that all the projects have the same CICD platform and feature store selections, which may not be the case.
For this I think we don't have to be too concerned about different CI/CD platform across projects since that's usually the point of a monorepo (and in general most companies only keep a single CI/CD platform across the whole company), but to address that and the feature store aspect, we can just ask them to regenerate the project and copy over only the workflow files (what we currently recommend).

mingyu89: Since there's already the 2-step for generating a different project in a monorepo, it's less likely for users to pause and check the additional readme before doing step2. This doc block will almost always be obsolete immediately.

arpitjasa-db: Yeah this is more for the data scientists who just initialized the project without CI/CD (not necessarily following the 2-step from the other project)

arpitjasa-db: Updated the README with the CICD/FS note!

template/{{.input_root_dir}}/README.md.tmpl
```diff
@@ -160,17 +160,28 @@ or pipeline assets (e.g. use a larger instance type for model training) via pull
...
-3. Copy yaml files from `project2/.azure/devops-pipelines/` to `monorepo_root_dir/.azure/devops-pipelines/` and make sure there's no name conflicts.
+2. Duplicate the YAML files in `monorepo_root_dir/.github/workflows/` that start with `project1-`, and be sure to change `project1` to `project2` for:
+    - File names
+    - The `'working-directory'` parameter in each file
+    - The `'pull_request'` `'paths'` parameter in the CI and tests workflow files
+{{- else if (eq .input_cicd_platform `azure_devops`) }}
+2. Duplicate the YAML files in `monorepo_root_dir/.azure/devops-pipelines/`, and be sure to change `project1` to `project2` for:
+    - File names
+    - The `'workingDirectory'` global variable in each file
+    - The `'trigger'` `'paths'` `'include'` parameter in each file
```
mingyu89, line 178: It's much more work and error prone to copy all the files, rename them and update all the parameters.
It's also additional context to learn to check if the configs for the new and existing projects are the same or not and then generate the project again.
Can we always generate the CICD files with correct parameters and let users do a copy and paste?

arpitjasa-db: Yeah two reasons for this:
1. Monorepo generation was failing in the same directory before and we have that open as a GitHub issue.
2. Data scientists often don't know CI/CD parameters anyway and now the UJ for them to get started is simpler.

The copy paste approach is the one we have currently and we've seen feedback that it's confusing. Long-term we were discussing separating the architecture into two DABs anyway where the MLEs set up the infra and the data scientists can just plug and play their own project with different templates (could also be multiple DABs tbd), and then plug into the MLE infrastructure when it's ready to productionize.

niall-turbitt: was going to make the same point about automating this for the user if they create a second project in the same repo. If it's possible for us to add the new project2 folder to the repo, could we not also add the 2 CI/CD yaml files under `devops-pipelines`, parameterized on that new project?

niall-turbitt: if we necessarily have to have the user duplicate the YAML files, I would also call out that they need to follow the same setup steps [here](https://github.com/databricks/mlops-stacks/blob/main/template/%7B%7B.input_root_dir%7D%7D/docs/mlops-setup.md.tmpl) to create new CI/CD pipelines for the new project

template/{{.input_root_dir}}/.azure/devops-pipelines/{{.input_project_name}}-bundle-cicd.yml.tmpl
niall-turbitt: if you're adding workingDirectory as another variable, the syntax needs to be changed here. It should be:

```
variables:
  - name: workingDirectory
     value: {{template `project_name_alphanumeric_underscore` .}}
  - group: {{template `project_name` .}} variable group 
```

Otherwise it results in the following error Screenshot 2024-01-11 at 11 09 03

template/{{.input_root_dir}}/README.md.tmpl
niall-turbitt: we would also want to call out in the MLOps docs [here](https://github.com/databricks/mlops-stacks/blob/main/template/%7B%7B.input_root_dir%7D%7D/docs/mlops-setup.md.tmpl#L192) that in the case of multiple projects, the build validation polices should be defined with a path filter to the respective project folders under the mono repo

template/{{.input_root_dir}}/.azure/devops-pipelines/{{.input_project_name}}-bundle-cicd.yml.tmpl
arpitjasa-db: Thanks for calling out, fixed! Anything else that needs to be changed for ADO?

template/{{.input_root_dir}}/README.md.tmpl
arpitjasa-db: Yeah we can call out that they need to follow the setup again for ADO and we can definitely add that build validation policy section to the docs

niall-turbitt: re the call out on path filters for build validation branch policies, let's replace [this line](https://github.com/databricks/mlops-stacks/blob/main/template/%7B%7B.input_root_dir%7D%7D/docs/mlops-setup.md.tmpl#L192) with the following:

```
1. Define two [build validation branch policies](https://learn.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops&tabs=browser#build-validation) for the `{{template `default_branch` .}}` branch using the two Azure build pipelines created in step 1. This is required so that any PR changes to the `{{template `default_branch` .}}` must build successfully before PRs can complete. In the case of a monorepo, where there are multiple projects under a single repository, set a [path filter](https://learn.microsoft.com/en-us/azure/devops/repos/git/branch-policies?view=azure-devops&tabs=browser#path-filters) on the build validation policies, such that devops pipelines are only triggered when there are changes to the respective projects (e.g. the path filter would be `/project1/*` to trigger a devops pipeline when changes are made to _only_ files under the `project1` folder)
```

template/{{.input_root_dir}}/.azure/devops-pipelines/{{.input_project_name}}-bundle-cicd.yml.tmpl
niall-turbitt: everything else looks good! Tested out on [this repo](https://dev.azure.com/niallturbitt/_git/mono-mlops-stacks)

template/{{.input_root_dir}}/README.md.tmpl
```diff
@@ -160,17 +160,29 @@ or pipeline assets (e.g. use a larger instance type for model training) via pull
...
+3. Follow the [Steps](./docs/mlops-setup.md#Steps) in docs/mlops-setup.md to make sure the pipelines are linked correctly in the ADO portal and the build validation branch policies are updated accordingly as well.
 {{ end -}}
 
+NOTE: If the second project uses a different CI/CD platform or feature store configuration, then the project will need to go through the full CI/CD setup process
```
vladimirk-db, line 182: This is important and should also go into the user-facing docs in databricks_template_schema.json IMO

template/{{.input_root_dir}}/README.md.tmpl
vladimirk-db: When we generate a monorepo in the first place, can we make the generated CI/CD workflows as generic as possible so they don't even need to be copy pasted later?? Or maybe the only update the DS needs to do is to just add paths for project2 to an existing CI/CD setup? (Looking at the generated files, I guess this is not feasible, but please correct me if I'm wrong).

vladimirk-db: > but the advantage here is that it addresses the feedback that DS usually don't have CI/CD information and keeps things simpler for them
 
Are we addressing the case where users are adding a new project to an existing repo? In that case CI/CD has already been setup for the previous projects though. I wonder if we can automate the setup in that case (and add a bunch of TODOs where it needs updating) for them.

Or the case where there is no CI/CD at all (in that case how would it be added later?)


Should the options be like: cicd_setup -> use existing/none/generate CICD workflows from scratch.

arpitjasa-db: So they can't just use existing natively since the CICD workflows are project-specific, but only based on the working directory. That's pretty much the only difference. 

We could generate the generic workflows but then we'd need to ask for `cicd_platform` at least as a parameter, and we'd have to assume `main` branch name, `release` branch name, `staging_workspace_host`, and `production_workspace_host`, marking as TODO for later on. Is that a better experience?

mingyu89: Interesting. We can completely decouple CICD from the project. The current problem is that a workflow is project-specific and needs manual modification. To resolve that, we can make the github/ado workflow file auto-generated from a more generic "workflow-generation workflow".

---

### 1. Create a data asset bundle template for generating CICD yml files for a specific project.
The template input variables are the following
- project_name
- cicd_platform
- main_branch
- release_branch
- staging_workspace_host
- production_workspace_host
- is_fs_enabled
- is_recipes_enabled

The output files are the CICD yaml files with project name in the prefix and proper files skipped based on input.

### 2. Add a yaml file in each project dir so that the "workflow-generation workflow" can load project_name, is_fs_enabled and is_recipes_enabled
The file can be something like: `<root_dir>/<project_name>/.proj_cicd_conf`
```
project_name : my_project
is_fs_enabled : yes
is_recipes_enabled : no
```

### 3. Add a yaml file in root dir so that the "workflow-generation workflow" can load CICD configs
The file can be something like: `<root_dir>/<project_name>/.root_cicd_conf`
```
cicd_platform : github
main_branch : main
release_branch : release
staging_workspace_host : www.staging.com
production_workspace_host : www.prod.com
```

### 4. Add the "workflow-generation workflow"
1. The workflow is triggered whenever there's a PR or commit.
2. read the `.root_cicd_conf`
3. loop through first level sub dirs and try to find the `.proj_cicd_conf`
4. For each project dir with `.proj_cicd_conf`, call bundle template to generate and workflow yml files and copy to the workflow dir.
5. If there's any update in the repo, commit and push the change.

---

Another benefit is that, now it's also possible to use a slightly different version of the CICD bundle template to add CICD to an existing project-only stacks prject.

arpitjasa-db: Yeah @sebastiaanboer-db and I were talking about this, but idk how it works if we go down the multiple templates route as of now. We'd have to figure out a few things (such as what initialization looks like, one point-of-entry switching to multiple, how this will be exposed to users via the CLI/UI, etc.) and there are a few related drawbacks to it as well.

In terms of the workflow you've proposed, we'd have to commit changes back to the repo as you said. So technically there would be a delay while the workflow-generation workflow runs, commits the changes back, then the project-specific workflow files run again (hopefully, not sure if committed workflows run retroactively on open PRs). Might be better to run this workflow only on PRs when one of the conf files is modified. But I'm not sure if this way is better, I think it does complicate things and requires us to create a separate template/repo

vladimirk-db: To clarify, we are basically addressing the case when DS is creating new projects in existing repo here, right? I like @mingyu89 's direction - I wonder if we can do something along those lines without revamping everything (yet).
For instance, we store the cicd conf on first initialization and reuse it later to automatically generate CICD workflow configs as new projects are added? We can discuss exactly how to setup this automation (this is all part of bundle init?)

(this CICD conf can become the basis for a separate template in the future)

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/assets/ml-artifacts-asset.yml.tmpl
```diff
@@ -1,56 +1,3 @@
...
-          comment: Registered model in Unity Catalog for the "{{template `project_name` .}}" ML Project for ${bundle.target} deployment target.{{end}}
-
-  prod:
-    resources:
-      {{ if (eq .input_include_models_in_unity_catalog `no`) }}models:
-        model:
-          description: |
-            MLflow registered model for the "{{template `project_name` .}}" ML Project. See the corresponding [Git repo]($#{var.git_repo_url}) for details on the project.
-
-            Links:
```
arpitjasa-db, line 40: Removing these links since `var.git_repo_url` has been broken for a while, and linking to these jobs does not merit providing `databricks_prod_workspace_host` as a variable.

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/requirements.txt.tmpl
```diff
@@ -1,8 +1,9 @@
 mlflow==2.7.1
 numpy>=1.23.0
-pandas>=1.4.3
+pandas>=1.5.3
```
arpitjasa-db, line 3: Updating these to match the root dir requirements

niall-turbitt:
testing this out and have taken the following workflow:
- create a project with cicd plus project (intending to make this a monorepo)
- add second project, using project_only option

All works fine at this stage. I have a mono repo with 2 projects and CI/CD for the first project. I now want to add the CI/CD pipelines for the second project, and run through initialization again, selecting CICD_only, and setting the input_root_dir to the monorepo project root dir
- I get an error at this point, that there is already a .gitignore file
image
- I try deleting this gitignore file and try the same setup, but this time get an error that the readme exists
image

What should be the intended workflow to add the CI/CD pipelines for a second project in a monorepo?

niall-turbitt:
I'm getting this `fs_param.json` file generated for projects, when feature store is not selected:
image

Can we remove this?

arpitjasa-db:
> I'm getting this `fs_param.json` file generated for projects, when feature store is not selected: image
> 
> Can we remove this?

Yeah good question, we could remove it and check for the presence/absence of the file instead in our workflows (but that does complicate the code a bit). Or we can rename is to `project_params.json`. What would be better?

niall-turbitt:
in the generated top level README, it looks like we're generated an additional `(` under the `Setting up CI/CD` section
image

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests.yml.tmpl
```diff
@@ -16,8 +18,6 @@ env:
   ARM_CLIENT_SECRET: {{`${{ secrets.STAGING_AZURE_SP_CLIENT_SECRET }}`}}
   {{- end }}
 
-concurrency: {{template `project_name` .}}-feature-training-integration-test-staging
```
mingyu89, line 19: Why do we remove the concurrency? Two PRs of the same project could run integration test in the same staging workspace and cause failures.

mingyu89 (commented):
This is really a huge improvement of the CUJs.

template/{{.input_root_dir}}/cicd_params.json.tmpl
```diff
@@ -0,0 +1,8 @@
+{
+  "input_root_dir": "{{ .input_root_dir }}",
+  "input_cloud": "{{ .input_cloud }}",
+  "input_cicd_platform": "{{ .input_cicd_platform }}",
```
mingyu89, line 4: We let user select the input_cloud during both the CICD/root_dir creation and the project creation.
- CICD: the cloud is used to in the cicd creation workflow and some other places
- project: project initialization uses the cicd platform to decide node type, workspace, etc.

We need to add a validation to make sure the could is the same between the parent and the project during the CICD creation workflow.

mingyu89: The easiest way is to add a new input vairable `input_project_cicd_platform` to `fs_param.json`(rename it to `project_param.json`) and check if the two input variables are the same.

niall-turbitt:
> > I'm getting this `fs_param.json` file generated for projects, when feature store is not selected: image
> > Can we remove this?
> 
> Yeah good question, we could remove it and check for the presence/absence of the file instead in our workflows (but that does complicate the code a bit). Or we can rename is to `project_params.json`. What would be better?

`project_params.json` seems like a good option

template/update_layout.tmpl
```diff
@@ -2,6 +2,25 @@
...
+    {{ skip (printf `%s/%s` $root_dir `.github`) }}
+    {{ skip (printf `%s/%s` $root_dir `docs`) }}
+    {{ skip (printf `%s/%s` $root_dir `.gitignore`) }}
+    {{ skip (printf `%s/%s` $root_dir `README.md`) }}
+    {{ skip (printf `%s/%s` $root_dir `test-requirements.txt`) }}
+    {{ skip (printf `%s/%s` $root_dir `cicd.tar.gz`) }}
+    {{ skip (printf `%s/%s` $root_dir `cicd_params.json`) }}
+{{ else if (eq .input_setup_cicd_and_project `CICD_Only`) }}
+    {{ skip (printf `%s/%s` $root_dir $project_name_alphanumeric_underscore) }}
+    {{ skip (printf `%s/%s/%s-%s` $root_dir `.github/workflows` .input_project_name `run-tests.yml`) }}
```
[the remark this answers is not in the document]
arpitjasa-db: We're not skipping the entire directory though, because we still need the top-level workflows to be generated like the deploy CI/CD workflow

template/update_layout.tmpl
```diff
@@ -2,6 +2,25 @@
...
+    {{ skip (printf `%s/%s/%s-%s` $root_dir `.github/workflows` .input_project_name `bundle-ci.yml`) }}
+    {{ skip (printf `%s/%s/%s-%s` $root_dir `.github/workflows` .input_project_name `bundle-cd-prod.yml`) }}
+    {{ skip (printf `%s/%s/%s-%s` $root_dir `.github/workflows` .input_project_name `bundle-cd-staging.yml`) }}
+    {{ skip (printf `%s/%s/%s-%s` $root_dir `.azure/devops-pipelines` .input_project_name `tests-ci.yml`) }}
```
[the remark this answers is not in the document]
arpitjasa-db: We're not skipping the entire directory though, because we still need the top-level workflows to be generated like the deploy CI/CD workflow

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-bundle-cd-staging.yml.tmpl
```diff
@@ -25,11 +25,11 @@ env:
 
 jobs:
   staging:
-    concurrency: {{template `project_name` .}}-staging-bundle-job
+    concurrency: {{ .input_project_name }}-staging-bundle-jo
```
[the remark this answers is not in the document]
arpitjasa-db: Yeah good catch I changed this to test the tar.gz generation and forgot to change back

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests-fs.yml.tmpl
```diff
@@ -1,65 +0,0 @@
-name: Feature and Training Integration Tests for {{template `project_name` .}}
```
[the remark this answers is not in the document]
arpitjasa-db: Yeah I didn't update the tar.gz yet. Was planning to do that once the PR was in a more final state to avoid having to generate over and over again. Once the workflow is deployed though this will be automatic so that will be nice

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests.yml.tmpl
arpitjasa-db: Oh I thought concurrency is to prevent different jobs in the same workflow from running together at the same time

template/{{.input_root_dir}}/cicd_params.json.tmpl
arpitjasa-db: You mean `input_cloud` right?

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-bundle-cd-staging.yml.tmpl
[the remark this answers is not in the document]
arpitjasa-db: Updated the tar.gz to include all the latest changes so should be good

template/{{.input_root_dir}}/README.md.tmpl
```diff
@@ -125,52 +128,59 @@ contained in the following files:
...
+├── .github                     <- Configuration folder for CI/CD using GitHub Actions. {{ if (eq .input_setup_cicd_and_project `CICD_and_Project`)}} The CI/CD workflows deploy ML assets defined in the `./assets/*` folder with databricks CLI bundles.{{ end }}
 {{ else if (eq .input_cicd_platform `azure_devops`) -}}
-├── .azure                      <- Configuration folder for CI/CD using Azure DevOps Pipelines. The CI/CD workflows deploy ML assets defined in the `./assets/*` folder with databricks CLI bundles.
+├── .azure                      <- Configuration folder for CI/CD using Azure DevOps Pipelines. {{ if (eq .input_setup_cicd_and_project `CICD_and_Project`)}} The CI/CD workflows deploy ML assets defined in the `./assets/*` folder with databricks CLI bundles.{{ end }}
 {{ end -}}
+|
+|── docs                        <- Contains documentation for the repo.
+|
+|── cicd.tar.gz                 <- Contains CI/CD bundle that should be deployed by {{.input_root_dir}}-deploy-cicd.yml to set up CI/CD for projects.
```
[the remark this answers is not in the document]
arpitjasa-db: Fixed

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests-fs.yml.tmpl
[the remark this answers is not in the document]
arpitjasa-db: Updated the tar.gz to include all the latest changes so should be good

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests.yml.tmpl
arpitjasa-db: Nvm looks like I misunderstood it, re-added it

template/{{.input_root_dir}}/cicd_params.json.tmpl
arpitjasa-db: Added a comparison check for `input_cloud`.

template/update_layout.tmpl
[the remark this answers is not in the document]
arpitjasa-db: Do you know if it's possible for us to define a regex like `input_project_name-*`? If so that would make our life a ton easier and we wouldn't have to specify each file cc @shreyas-goenka

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests-fs.yml.tmpl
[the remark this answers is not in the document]
arpitjasa-db: Jk looks like the workflow ran woohoo! We're good

template/{{.input_root_dir}}/README.md.tmpl
```diff
@@ -15,22 +15,25 @@ This project contains the following components:
...
+{{ end }}
 | CI/CD                      | {{ if (eq .input_cicd_platform `github_actions`) }}[GitHub Actions](https://github.com/actions) workflows to test and deploy ML code and assets {{ else if (eq .input_cicd_platform `azure_devops`) }}[Azure DevOps Pipelines](https://azure.microsoft.com/en-gb/products/devops/pipelines/) to test and deploy ML code and assets{{ end }}      |
 
 contained in the following files:
 
 ```
-{{ template `root_dir` .}}        <- Root directory. Both monorepo and polyrepo are supported.
+{{ .input_root_dir }}        <- Root directory. Both monorepo and polyrepo are supported.
 │
+{{ if (eq .input_setup_cicd_and_project `CICD_and_Project`)}}
```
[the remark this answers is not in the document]
mingyu89: With github.
The component table is also not correct.
Screenshot 2024-02-02 at 2 02 37 AM

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
```diff
@@ -24,6 +24,7 @@ targets:
     workspace:
       # TODO: add dev workspace URL
       host:
+{{ if (eq .input_setup_cicd_and_project `CICD_and_Project`)}}
```
mingyu89, line 27: Why are we removing the test, staging and prod? 
After setting up CICD, users will have to add the block themselves.

It's better to add a dummy host name & comments than removing the block.
The token can be overwritten in the env var. I think the host can be overwritten too.
In this case users won't need additional step after running the cicd deployment workflow.

template/{{.input_root_dir}}/README.md.tmpl
```diff
@@ -124,53 +127,60 @@ contained in the following files:
...
-{{ if or (eq .input_cicd_platform `github_actions`) (eq .input_cicd_platform `github_actions_for_github_enterprise_servers`) -}}
-3. Copy yaml files from `project2/.github/workflows/` to `monorepo_root_dir/.github/workflows/` and make sure there's no name conflicts.
+| Data Scientist                | Modify production model ML assets, e.g. model training or inference jobs  | {{ if (eq .input_setup_cicd_and_project `CICD_and_Project`)}}[ML asset config guide]({{template `project_name_alphanumeric_underscore` .}}/assets/README.md) {{ else }} ML assets README when project is initialized {{ end }} |
+
+## Setting up CI/CD
+This stack comes with a workflow to set up CI/CD for projects that can be found in
+{{ if (eq .input_cicd_platform `azure_devops`)}}
+`.azure/devops-pipelines/deploy-cicd.yml`.
+{{ else }}
+`.github/workflows/deploy-cicd.yml`.
```
mingyu89, line 176: Nit: the `.github/workflows/deploy-cicd.yml` or `.azure/devops-pipelines/deploy-cicd.yml` is on a different line
Screenshot 2024-02-02 at 2 18 44 AM

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
arpitjasa-db: Actually in the deploy CI/CD workflow, we modify `databricks.yml` so they don't have to add it in themselves

mingyu89: Oh I see. It depends on the assumption that users won't modify the `databricks.yml` and make it incompatible. For example, if user removes the empty lines or add another element after `targets`, the logic will be broken.
Let's also create a ticket to follow up and not modify the project `databricks.yml`.

mingyu89: Sometime in the future, we may be able to use LLM to analysis the `databricks.yml` file and safely make the updates. Until then, the best way is to not modify it, or have a reserved placeholder block for auto generated code.

template/{{.input_root_dir}}/.github/workflows/{{.input_project_name}}-run-tests.yml.tmpl
mingyu89: 👍

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
arpitjasa-db: Follow-up or just update the repo now to have dummy URLs instead?

template/{{.input_root_dir}}/README.md.tmpl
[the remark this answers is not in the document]
arpitjasa-db: Fixed!

template/{{.input_root_dir}}/README.md.tmpl
```diff
@@ -124,53 +127,60 @@ contained in the following files:
...
+{{ end }}
+
+To set up CI/CD for projects that were created through MLOps Stacks with the `Project_Only` parameter, 
+run the abovementioned workflow, specifying the `project_name` as a parameter. This workflow assumes that 
+all steps in the [MLOps Setup Guide](./docs/mlops-setup.md#Steps) have been completed.
```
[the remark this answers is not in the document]
arpitjasa-db: Rendered for me, do you want to try again? Just updated the code a bit, it could be the old issue that Niall pointed out

template/{{.input_root_dir}}/test-requirements.txt.tmpl
```diff
@@ -1,2 +1,2 @@
 pytest>=7.1.2
-pandas==1.5.3
```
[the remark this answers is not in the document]
arpitjasa-db: Filed https://databricks.atlassian.net/browse/MLOPS-612

library/template_variables.tmpl
```diff
@@ -1,58 +1,4 @@
 # define template variables
-{{ define `root_dir` -}}
-    {{ .input_root_dir }}
-{{- end }}
-
-{{ define `project_name` -}}
-    {{ .input_project_name }}
```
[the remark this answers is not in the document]
arpitjasa-db: Fixed!

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
mingyu89: It's fine for this PR.
Adding the `host` to env var will need a manual test to make sure it works as I'm not 100% confident.
Having a working version is more important as it's already huge.

mingyu89:
If I approve, please merge the PR shortly before the bugbash so that we can catch issues and fix them quickly.

arpitjasa-db:
> If I approve, please merge the PR shortly before the bugbash so that we can catch issues and fix them quickly.

Sure I can hold off on merging this till Monday morning, but I do also want to get the `resources` rename PR in before the bugbash

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
arpitjasa-db: Filed https://databricks.atlassian.net/browse/MLOPS-613

.github/workflows/generate-cicd-zip.yaml
```diff
@@ -0,0 +1,36 @@
+name: Generate CICD Zip
```
mingyu89, line 1: For the issue you mentioned, two options
1. Put both in the same workflow.
2. add the same `concurrency` group to both workflows so that one of them will be queued.

template/update_layout.tmpl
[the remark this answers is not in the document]
shreyas-goenka: @arpitjasa-db Yeah we support glob patterns for the file name, you don't need to pass in all file paths!

You can see https://pkg.go.dev/path#Match for the glob pattern syntax supported.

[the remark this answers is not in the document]
arpitjasa-db: Perfect ty! Updated

mingyu89 (approved):

niall-turbitt:
couple of minor things that need to be fixed in order to get the deploy cicd pipeline running for Azure DevOps. Aside from those, looks good. Thanks for this @arpitjasa-db , appreciate this was a tricky one to work through!

Set up a mono repo to test here https://dev.azure.com/niallturbitt/_git/monorepo-mlops-stacks - have added you as an admin Arpit

niall-turbitt:
just catching this. For the second project I generated there were no staging or prod profiles defined in the `databricks.yaml` file. Understand the reasoning for this; a data scientist adding a second project may not know up front the staging/prod workspace URLs.

image

What this means though, is that when the CI/CD pipelines are triggered for the second project, there is not staging/prod profiles defined for the bundles CI/CD pipeline to use

image

We either need to somehow automate the addition of staging/test profiles to the databricks.yaml file of the new project whenever we are running the deploy cicd pipeline for the new project. Or direct the user to do so somewhere in documentation

arpitjasa-db:
> just catching this. For the second project I generated there were no staging or prod profiles defined in the `databricks.yaml` file. Understand the reasoning for this; a data scientist adding a second project may not know up front the staging/prod workspace URLs.
> 
> 
> 
> image
> 
> 
> 
> What this means though, is that when the CI/CD pipelines are triggered for the second project, there is not staging/prod profiles defined for the bundles CI/CD pipeline to use
> 
> 
> 
> image
> 
> 
> 
> We either need to somehow automate the addition of staging/test profiles to the databricks.yaml file of the new project whenever we are running the deploy cicd pipeline for the new project. Or direct the user to do so somewhere in documentation


@niall-turbitt I actually added the profile generation as part of the deploy CICD pipeline. Did it not work?

arpitjasa-db:
> @niall-turbitt I actually added the profile generation as part of the deploy CICD pipeline. Did it not work?

Ah actually I see we forgot to include that file when we did `git add`. It should be fixed now! Also made the other changes you had in the monorepo I think. Would you mind taking a final pass @niall-turbitt and see if all the issues you found were fixed? Thank you!

niall-turbitt:
@arpitjasa-db the above comments I had added but didn't properly publish, so don't think you caught them. I see the `refs/heads/feature-branch-pr` branch issue has already been resolved.

The only other thing I think needs addressing is [this point](https://github.com/databricks/mlops-stacks/pull/133#discussion_r1477382846) around permissions for the build service to be able to run the deploy CICD pipeline successfully

niall-turbitt:
the only other minor nit that I think could be confusing for users is the following in a monorepo setup:

- I am adding a second project to a monorepo, and I specify the project name as `project-2` during initialization
- This is added under a folder `project_2` in the monorepo
- When running the deploy CICD pipeline to set up the CICD pipelines for project 2 we have to set the name of the project via a param:
```
parameters:
  - name: ProjectName
    displayName: Name of the project to deploy CI/CD for.
    default: mono-mlops-stacks
    type: string
```

From the default `mono-mlops-stacks` I would assume this param should be `project-2` to set up for the new project. However this param has to be `project_2` to map to the folder

template/{{.input_root_dir}}/.azure/devops-pipelines/deploy-cicd.yml.tmpl
```diff
@@ -0,0 +1,93 @@
...
+        git checkout -b add-cicd-for-{{`${{ parameters.ProjectName }}`}}
+        git add .azure
+        git commit -m "Add CICD for {{`${{ parameters.ProjectName }}`}}"
+        git push origin add-cicd-for-{{`${{ parameters.ProjectName }}`}}
+      displayName: Push CICD Bundle to a New Branch
+
+    # Create Pull Request
+    - script: |
+        echo Creating Pull Request...
+        PR_DATA="{\"sourceRefName\":\"refs/heads/feature-branch-pr\",\"targetRefName\":\"refs/heads/main\",\"title\":\"Merging Feature Branch\",\"description\":\"Pull request from feature branch\",\"reviewers\":[]}"
```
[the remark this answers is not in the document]
niall-turbitt: resolved

template/{{.input_root_dir}}/.azure/devops-pipelines/deploy-cicd.yml.tmpl
```diff
@@ -0,0 +1,93 @@
...
+
+    # Create Branch and Commit CICD Bundle
+    - script: |
+        git config --global user.name "Deploy CICD Bot"
+        git config --global user.email "noreply-cicd-bot@databricks.com"
+        git checkout -b add-cicd-for-{{`${{ parameters.ProjectName }}`}}
+        git add .azure
+        git commit -m "Add CICD for {{`${{ parameters.ProjectName }}`}}"
+        git push origin add-cicd-for-{{`${{ parameters.ProjectName }}`}}
+      displayName: Push CICD Bundle to a New Branch
```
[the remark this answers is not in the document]
arpitjasa-db: Bit concerned about adding `--force` on the off-chance someone has the same branch or has force pushes disabled on their repository. Seems a bit weird to need to do add this flag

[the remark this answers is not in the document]
arpitjasa-db: Added `--force-with-lease` to prevent the overwriting possibility, and someone can just update the pipeline to remove this if they have force disabled on the whole repo/org

arpitjasa-db:
> one other minor nit that I think could be confusing for users is the following in a monorepo setup:
> 
> * I am adding a second project to a monorepo, and I specify the project name as `project-2` during initialization
> * This is added under a folder `project_2` in the monorepo
> * When running the deploy CICD pipeline to set up the CICD pipelines for project 2 we have to set the name of the project via a param:
> 
> ```
> parameters:
>   - name: ProjectName
>     displayName: Name of the project to deploy CI/CD for.
>     default: mono-mlops-stacks
>     type: string
> ```
> 
> From the default `mono-mlops-stacks` I would assume this param should be `project-2` to set up for the new project. However this param has to be `project_2` to map to the folder

@niall-turbitt good catch! Added a step in GH and ADO to create a underscore version of the project name and use that for the directory instead so they can continue using the normal name

PR #147, opened 2024-02-07
Fix README formatting & shorten initialization prompts

arpitjasa-db:
We shorten the initialization prompts to less than 80 characters to prevent line duplication issue from occurring in the terminal. This line overflow may still occur if the terminal window is shortened to less than 80 chars per line, but CLI team has been informed and plan to fix this by the end of the quarter, so this is a short-term mitigation.

### DIFF
--- a/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
+++ b/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
@@ -37,7 +37,7 @@ contained in the following files:
 │   │
 │   ├── databricks.yml          <- databricks.yml is the root bundle file for the ML project that can be loaded by databricks CLI bundles. It defines the bundle name, workspace URL and resource config component to be included.
 │   │
-{{- if and (eq .input_include_feature_store `no`) (eq .input_include_mlflow_recipes `no`) -}}
+{{- if and (eq .input_include_feature_store `no`) (eq .input_include_mlflow_recipes `no`) }}
 │   ├── training                <- Training folder contains Notebook that trains and registers the model.
 │   │
 │   ├── validation              <- Optional model validation step before deploying a model.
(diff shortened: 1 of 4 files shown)


template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
```diff
@@ -37,7 +37,7 @@ contained in the following files:
 │   │
 │   ├── databricks.yml          <- databricks.yml is the root bundle file for the ML project that can be loaded by databricks CLI bundles. It defines the bundle name, workspace URL and resource config component to be included.
 │   │
-{{ if and (eq .input_include_feature_store `no`) (eq .input_include_mlflow_recipes `no`) -}}
+{{- if and (eq .input_include_feature_store `no`) (eq .input_include_mlflow_recipes `no`) -}}
```
mingyu89, line 40: Could you try generating the file?
I suspect it will generate something like:
```
│ ││ ├── training <- Training folder contains Notebook
```

mingyu89: The original state is 
```
{{ if .. -}}
...
{{ else if .. -}}
...
{{ else -}}
...
{{ end -}}
```
Every condition is on the separate line(same as current state).

arpitjasa-db: Yup great call out, removed the following `-` to ensure the space doesn't get deleted so it looks better now 
Screenshot 2024-02-07 at 12 41 13 PM

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
```diff
@@ -37,7 +37,7 @@ contained in the following files:
 │   │
 │   ├── databricks.yml          <- databricks.yml is the root bundle file for the ML project that can be loaded by databricks CLI bundles. It defines the bundle name, workspace URL and resource config component to be included.
 │   │
-{{ if and (eq .input_include_feature_store `no`) (eq .input_include_mlflow_recipes `no`) -}}
+{{- if and (eq .input_include_feature_store `no`) (eq .input_include_mlflow_recipes `no`) }}
```
mingyu89, line 40: I think this file works correctly before the change.
It's still correct after the change.

mingyu89 (approved):

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/README.md.tmpl
arpitjasa-db: Gotcha makes sense, but yeah consistency

PR #155, opened 2024-04-12
Update the version of matplotlib to avoid security issues

lbcommer:
Update the version of matplotlib to avoid security issues with the version of its required package Pillow. This way, the installed version of Pillow will be at least 10.3.0, which doesn't have the problematic security issue: https://security.snyk.io/vuln/SNYK-PYTHON-PILLOW-5918878

### DIFF
--- a/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/requirements.txt.tmpl
+++ b/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/requirements.txt.tmpl
@@ -2,7 +2,8 @@ mlflow==2.7.1
 numpy>=1.23.0
 pandas==1.5.3
 scikit-learn>=1.1.1
-matplotlib>=3.8.4
+matplotlib>=3.5.2
+pillow>=10.0.1
 Jinja2==3.0.3
 pyspark~=3.3.0
 pytz~=2022.2.1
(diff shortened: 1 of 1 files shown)


arpitjasa-db:
Thanks for the PR @lbcommer! It seems though that we're running into some version dependency and incompatibility issues with a naive upgrade. This can be something we look into and update all our dependencies accordingly

lbcommer:
@arpitjasa-db This security vulnerability is scored as critical. Any new project created with the template will introduce the security risk. I think it makes sense to apply a quick solution. That's not against any other further improvement/update of the rest of the dependencies.

arpitjasa-db:
@lbcommer yeah I understand, but what I meant was this upgrade seems to break MLOps Stacks due to incompatible dependencies not resolving correctly, as we see from the above checks that ran.

lbcommer:
@arpitjasa-db I see the problem. I modified the PR to include in the requirements the `pillow` package instead, since a version without the security issue (10.0.1). This way, all tests have passed. The last `matplotlib` works with `pillow>=8` (https://github.com/matplotlib/matplotlib/blob/cbdd3b373ecf6074a8813465c16721d889852642/environment.yml#L22) and `matplotllib` 3.5.2 required `pillow>=6.2` so, all should be ok.

arpitjasa-db (approved):

arpitjasa-db:
@lbcommer fantastic thanks a bunch for making this change!

PR #160, opened 2024-06-06
minor change to deploy.py.tmpl to fix the duplicated alias assigned to the created model in UC

yinxi-db:
change all alias to lower case, mlflow UC model registry does not different upper/lower case in alias. The original code leads to model in UC have both alias "champion" and "challenger"
image

### DIFF
--- a/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/deployment/model_deployment/deploy.py.tmpl
+++ b/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/deployment/model_deployment/deploy.py.tmpl
@@ -41,7 +41,7 @@ def deploy(model_uri, env):
     _, model_name, version = model_uri.split("/")
     client = MlflowClient(registry_uri="databricks-uc")
     mv = client.get_model_version(model_name, version)
-    target_alias = "Champion"
+    target_alias = "champion"
     if target_alias not in mv.aliases:
         client.set_registered_model_alias(
             name=model_name,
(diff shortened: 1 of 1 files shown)


arpitjasa-db:
@yinxi-db bit confused, if aliases are case-insensitive, wouldn't this change be a no-op? Have we confirmed this works?

yinxi-db:
> @yinxi-db bit confused, if aliases are case-insensitive, wouldn't this change be a no-op? Have we confirmed this works?

yes, see the registered model [here](https://adb-984752964297111.11.azuredatabricks.net/explore/data/models/test/dais_mlops_demo/dais_mlops_demo-model?o=984752964297111). Without the change, it will create a version with 2 alias as shown in the screenshot earlier and with the update, version 3 has the alias `champion` since it passed validation

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/deployment/model_deployment/deploy.py.tmpl
```diff
@@ -41,20 +41,20 @@ def deploy(model_uri, env):
     _, model_name, version = model_uri.split("/")
     client = MlflowClient(registry_uri="databricks-uc")
     mv = client.get_model_version(model_name, version)
-    target_alias = "Champion"
+    target_alias = "champion"
```
arpitjasa-db, line 44: Can we change the other places in the Stack as well where use `Champion` or `Challenger` like the batch inference and model validation notebooks?

arpitjasa-db:
This was done as part of https://github.com/databricks/mlops-stacks/pull/165

Issue #70, opened 2023-03-20
Model validation with feature store should be supported.


mingyu89:
Mode validation uses `mlflow.evaluate`. Currently it doesn't support evaluating models registered with databricks feature store.

aminebizid:
Any news about this limitation please?

jackroseman:
I am also wondering when this feature is coming. Thanks!

arpitjasa-db:
Hi we're looking into adding support for this, the tricky part is we need to update underlying feature store and MLflow components to get this to work, but will hopefully have an update for you all soon!

david-straub:
Any updates?

arpitjasa-db:
@david-straub we are currently reviewing a PR to add this functionality, but note this workaround will be slightly limited in that we cannot enable baseline comparison, as we need to provide a model_uri of a pyfunc model (so this means it's not possible to compare a new FS model with a baseline FS model)

arpitjasa-db:
We just merged https://github.com/databricks/mlops-stacks/pull/165 thanks to @aliazzzdat! This workaround should resolve most use cases for this issue, but if there's any other problems, please open another issue thank you!

PR #165, opened 2024-07-17
Model validation for FS models

aliazzzdat:
FS models cannot predict as pyfunc models yet (issue #70), thus we can't use MLflow Evaluate with FS models.

However, MLflow Evaluate can take a function instead of a pyfunc model_uri. 
This means we can call an FS model in the function and use this function in the Evaluate method.

Although this is a good solution, we still cannot enable baseline comparison, as we need to provide a model_uri of a pyfunc model (so this means it's not possible to compare a new FS model with a baseline FS model).


arpitjasa-db:
@aliazzzdat thanks for all the changes! Looks like some tests are failing, can we fix those and also do a full instantiation of Stacks to confirm everything is still working as expected? Version changes can be tricky, and want to make sure we don't break anything

tests/example-project-configs/aws/aws-github.json
```diff
@@ -4,12 +4,12 @@
   "input_project_name": "example-aws-github-project",
   "input_cloud": "aws",
   "input_cicd_platform": "github_actions",
-  "input_databricks_staging_workspace_host": "https://your-staging-workspace.cloud.databricks.com",
-  "input_databricks_prod_workspace_host": "https://your-prod-workspace.cloud.databricks.com",
+  "input_databricks_staging_workspace_host": "https://e2-demo-field-eng.cloud.databricks.com/",
```
arpitjasa-db, line 7: Revert changes to this file please

aliazzzdat:
The batch inference works as expected once we create the feature_store_inference_input table 
Screenshot 2024-07-19 at 21 54 34

.gitignore
```diff
@@ -11,4 +11,4 @@
 __pycache__/
 .cache
 *.pyc
-mlops-stacks.iml
+mlops-stacks.iml
```
arpitjasa-db, line 14: nit newline

tests/example-project-configs/aws/aws-github-fs.json
```diff
@@ -0,0 +1,18 @@
+{
```
arpitjasa-db, line 1: What's the purpose of this file? We don't run tests using this right?

aliazzzdat: It's just the file I used to generate the template to test whether the changes were working. It's the same as the other one, except it generates an FS model.

.gitignore
aliazzzdat: I added and removed  a file in my local .gitignore, the changes are just the space removed after .iml
Should we add *.iml in the file or do you want me revert the changes (adding the space back) ?

arpitjasa-db: Let's revert the changes

tests/example-project-configs/aws/aws-github-fs.json
arpitjasa-db: Hmm we don't really get much utility out of it except specify FS as yes, which if we did for each option, end up with tens of example configs. I think we should just remove this file then

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/validation/notebooks/ModelValidation.py.tmpl
```diff
@@ -255,8 +257,83 @@ def log_to_model_description(run, success):
...
+
+data = rounded_taxi_data(data)
+
+
+{{ end }}
+
 # COMMAND ----------
 
+{{ if (eq .input_include_feature_store `yes`) }}
+###########################################################################################
```
arpitjasa-db, line 315: Why do we have these lines of `#`?

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/validation/notebooks/ModelValidation.py.tmpl
```diff
@@ -277,7 +354,11 @@ with mlflow.start_run(
 
     try:
         eval_result = mlflow.evaluate(
+            {{ if (eq .input_include_feature_store `yes`) }}
+            get_fs_model,
```
arpitjasa-db, line 358:
```suggestion
            model=get_fs_model,
```

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/validation/notebooks/ModelValidation.py.tmpl
aliazzzdat: yes it's leftover comments that I put in place when I showed the fix to the FS team, I'm going to remove them

arpitjasa-db (approved):
Thanks for all the changes @aliazzzdat !

PR #170, opened 2024-09-04
Add metadata file to link to CLI deploy

arpitjasa-db:


template/{{.input_root_dir}}/.stacks_metadata.yml.tmpl
```diff
@@ -0,0 +1,11 @@
+# DO NOT MODIFY THIS AUTO-GENERATED FILE. 
+# MODIFYING THIS FILE MAY CAUSE YOUR STACK TO NOT FUNCTION PROPERLY
```
shreyas-goenka, line 2: Should we really claim this here?

arpitjasa-db: It is true, in the future we will use this UUID for other things too, so want to make sure no one modifies it

template/{{.input_root_dir}}/.stacks_metadata.yml.tmpl
```diff
@@ -0,0 +1,11 @@
...
+uuid: "{{ uuid }}"
```
shreyas-goenka, line 3: Idea: We could emit all these details as usage logs during template creation. Then we'll only have to persist:
```
uuid: <>
source_template: mlops-stacks
```

arpitjasa-db: Would that be possible since `bundle init` doesn't require authentication right? Would we use the unauthenticated usage log endpoint for this? Also, we'd need to still attach UUID to that initial log during template creation

shreyas-goenka: Does mlops-stacks use the `{{workspace_host}}` helper? If so auth is required. It might be required even if you don't use the helper. I'd have to check.

arpitjasa-db: We don't use the helper (we ask the user to specifically input workspaces that they plan to deploy to)

arpitjasa-db:
Closing for https://github.com/databricks/mlops-stacks/pull/170

PR #174, opened 2024-10-18
adding Gitlab support

valx:
Main changes:
- added support for Gitlab. The Gitlab pipeline files are in the folder `template/{{.input_root_dir}}/.gitlab/`  . Changes include:
   - Modified `/mlops-stacks/databricks_template_schema.json` prompt metadata by adding gitlab option
    - Searched all occurrences of `input_cicd_platform` and added a gitlab option
- in databricks bundle file, added `mode: development` to dev target


valx:
Need to add documentation in `/template/{{.input_root_dir}}/docs/mlops-setup.md.tmpl`

template/{{.input_root_dir}}/.gitlab/cd_production.yml.tmpl
```diff
@@ -0,0 +1,15 @@
+image: {{ .input_docker_image }}:latest
```
shichengzhou-db, line 1: qq, why are these yml.tmpl name convention not consistent with what's in .github and .azure?
e.g in .github this is called
.github/workflows/{{.input_project_name}}-bundle-cd-prod.yml.tmpl

README.md
```diff
@@ -79,9 +79,10 @@ This will prompt for parameters for initialization. Some of these parameters are
  * ``input_cloud``: Cloud provider you use with Databricks (AWS, Azure, or GCP).
 
 Others must be correctly specified for CI/CD to work:
- * ``input_cicd_platform`` : CI/CD platform of choice (GitHub Actions or GitHub Actions for GitHub Enterprise Servers or Azure DevOps)
- * ``input_databricks_staging_workspace_host``: URL of staging Databricks workspace, used to run CI tests on PRs and preview config changes before they're deployed to production.
-   We encourage granting data scientists working on the current ML project non-admin (read) access to this workspace,
+ * ``input_cicd_platform`` : CI/CD platform of choice (GitHub Actions or GitHub Actions for GitHub Enterprise Servers or Azure DevOps or Gitlab)
+ * ``input_databricks_test_workspace_host``: URL of staging Databricks workspace, used to run CI tests on PRs before they're deployed to staging.
```
arpitjasa-db, line 83: Do we need this? `input_databricks_test_workspace_host` is for dev or staging? If it's the same URL as staging we don't need another parameter right?

README.md
```diff
@@ -79,9 +79,10 @@ This will prompt for parameters for initialization. Some of these parameters are
  * ``input_cloud``: Cloud provider you use with Databricks (AWS, Azure, or GCP).
 
 Others must be correctly specified for CI/CD to work:
- * ``input_cicd_platform`` : CI/CD platform of choice (GitHub Actions or GitHub Actions for GitHub Enterprise Servers or Azure DevOps)
- * ``input_databricks_staging_workspace_host``: URL of staging Databricks workspace, used to run CI tests on PRs and preview config changes before they're deployed to production.
-   We encourage granting data scientists working on the current ML project non-admin (read) access to this workspace,
+ * ``input_cicd_platform`` : CI/CD platform of choice (GitHub Actions or GitHub Actions for GitHub Enterprise Servers or Azure DevOps or Gitlab)
```
arpitjasa-db, line 82:
```suggestion
 * ``input_cicd_platform`` : CI/CD platform of choice. Currently we support GitHub Actions, GitHub Actions for GitHub Enterprise Servers, Azure DevOps and GitLab.
```

databricks_template_schema.json
```diff
@@ -316,6 +331,44 @@
           }
         ]
       }
+    },      
+    "input_docker_image": {
+      "order": 20,
+      "type": "string",
+      "description": "\nDocker image for the execution of Gitlab pipelines",
+      "default": "registry.gitlab.com/ TODO /",
```
arpitjasa-db, line 339: What is the TODO for here?

databricks_template_schema.json
```diff
@@ -316,6 +331,44 @@
           }
         ]
       }
+    },      
+    "input_docker_image": {
```
arpitjasa-db, line 335: Does this need to be an initialization parameter (generally we set this when it's used across several files) or can this be a different kind of parameter (job parameter, manually copy-pasted, etc.)?

template/{{.input_root_dir}}/.gitlab/cd_production.yml.tmpl
arpitjasa-db: +1 it would really help if we can follow similar naming conventions across providers

arpitjasa-db (commented):
Thanks for the PR @valx ! Left a few comments, would you mind also generating a couple of Stacks projects from this PR and linking them in the PR? Would be good to verify this works with GitLab as expected, and the docs all look ok, as well as other providers remain unaffected.

README.md
valx: I have been working and developing this PR in a scenario with 4 environments (dev, integration test, staging, prod).

The parameter gives you the option to have a separate integration test environment, while you can still choose to use the same workspace as in staging, in case you do not have/need a separate workspace.

databricks_template_schema.json
valx: I assume that the user creates a Docker image with all the required libraries to run the unit tests and pushes it in their own or in a public registry.
(I need to document it)

databricks_template_schema.json
valx: good catch, I will probably remove it as a parameter and document that it needs to be replaced later by the user

template/{{.input_root_dir}}/.gitlab/cd_production.yml.tmpl
valx: updated to follow the naming conventions

README.md
valx: NOTE that I am using the integration test environment in the gitlab pipelines only.
so we either need to:
- modify Github/Azure pipelines to use it
- make this parameter restricted to the gitlab option only

databricks_template_schema.json
valx: The parameter is actually used in all the gitlab pipelines files.
The docker image is the image for a VM to run the CI/CD with all the required libraries . That is a best practice in gitlab as opposite to adding the commands to install the dependencies in the pipelines themselves, for the following reasons:
- you do not need to re-install the dependencies for each pipeline execution
- installing the dependencies on-the-fly in gitlab pipeline seems more complicated as you cannot leverage on tools like github actions/checkout

valx: I've updated the file
` template/{{.input_root_dir}}/docs/mlops-setup.md.tmpl`
 to describe the process to create and push the docker image

valx:
@arpitjasa-db 
as suggested, I have created 2 repos with projects generated from the template:
[https://github.com/valx/mlops-ex1](https://github.com/valx/mlops-ex1) (gitlab)
[https://github.com/valx/mlops-ex1](https://github.com/valx/mlops-ex2) (github)

databricks_template_schema.json
valx: I've updated the file
 template/{{.input_root_dir}}/docs/mlops-setup.md.tmpl
to describe the process to create and push the docker image

README.md
arpitjasa-db: Got it, generally we try to minimize the number of parameters so users don't get overwhelmed with the number of knobs, and it seems that this parameter is only used in one spot (`databricks.yml`). I think it would be better to add docs on how to change this if they want, but not add an extra parameter just for this one spot especially if in most cases it would stay the same as staging

databricks_template_schema.json
arpitjasa-db: Got it, but in most cases, we're still recommending to the user to provide a default TODO value at time of initialization since they won't have the image ready? At that point, they'd have to manually update all the spots with the image any way right? So wouldn't it be simpler to mark those spots with a TODO instead?

arpitjasa-db:
@valx any update on this PR? Looks like the tests are still failing

README.md
valx: removed the parameter

databricks_template_schema.json
valx: I agree if we assume they wouldn't have an image ready. 
But they might have the image ready, especially after they run their first project, they are likely to be able to reuse the same image.

valx:
> @valx any update on this PR? Looks like the tests are still failing

Done some changes, please allow the test to re-run

databricks_template_schema.json
arpitjasa-db: Publicly available image as default would be great! And in that case, we can leave as a parameter since it will only show up as an extra knob for GitLab but can still work right out of the box

template/{{.input_root_dir}}/docs/mlops-setup.md.tmpl
```diff
@@ -268,6 +269,87 @@ In the case of a monorepo, where there are multiple projects under a single repo
...
+Service principals can be created and managed in the your cloud provider identity solution or in Databricks directly. We normally recommend setting up a Databricks managed service principal.  
+Follow [Add a service principal to your Databricks account]({{ template `generate_doc_link` (map (pair "cloud" .input_cloud) (pair "path" "administration-guide/users-groups/service-principals.html#add-a-service-principal-to-your-databricks-account")) }})
+and [Add a service principal to a workspace]({{ template `generate_doc_link` (map (pair "cloud" .input_cloud) (pair "path" "administration-guide/users-groups/service-principals.html#add-a-service-principal-to-a-workspace")) }})
+for details.
+
+{{ if eq .input_cloud `aws` }}
+For your convenience, we also have a [Terraform module](https://registry.terraform.io/modules/databricks/mlops-aws-project/databricks/latest) that can set up your service principals.
+{{ end }}
+
+#### Configure Service Principal (SP) permissions
```
arpitjasa-db, line 298: Hmm seems like we're copy-pasting a lot of shared information, can we re-organize to minimize duplication? If that ends up being a heavy lift, then we can do that in a future PR since this is large enough as is

tests/test_gitlab.py
```diff
@@ -0,0 +1,40 @@
...
+    parametrize_by_cloud,
+)
+
+@pytest.mark.parametrize(
+    "cicd_platform", ["gitlab"]
+)
+@pytest.mark.parametrize(
+    "setup_cicd_and_project,include_feature_store,include_mlflow_recipes,include_models_in_unity_catalog",
+    [
+        ("CICD_and_Project", "no", "no", "yes"),
```
arpitjasa-db, line 16: Do we only allow GitLab with Models in UC?

tests/utils.py
```diff
@@ -138,6 +138,8 @@ def markdown_checker_configs(tmpdir):
             {"pattern": "http://127.0.0.1:5000"},
             {"pattern": "https://adb-3214.67.azuredatabricks.net*"},
             {"pattern": "https://adb-345.89.azuredatabricks.net*"},
+            {"pattern": "https://adb-xxxx.xx.azuredatabricks.net*"},
```
arpitjasa-db, line 141: Why are we adding these in?

tests/test_gitlab.py
valx: I've only tested it with models in UC. But there is nothing stopping us from allowing all cases. shall I copy all the option we use for github?

tests/utils.py
valx: It was originally for the test workspace, now removed. So I will remove this too.

tests/test_gitlab.py
valx: added all cases

template/{{.input_root_dir}}/docs/mlops-setup.md.tmpl
valx: I  copied the same text that was already copied between Azure and Github...

valx: I suggest to do it in a future PR, since the re-org should include all options, not only the changes relative to gitlab

databricks_template_schema.json
arpitjasa-db: Hmm this seems like a temporary solution, I'm ok with it if we can come up with a long-term solution for this to be a Databricks hosted image

tests/utils.py
```diff
@@ -138,6 +138,7 @@ def markdown_checker_configs(tmpdir):
             {"pattern": "http://127.0.0.1:5000"},
             {"pattern": "https://adb-3214.67.azuredatabricks.net*"},
             {"pattern": "https://adb-345.89.azuredatabricks.net*"},
+            {"pattern": "https://your-test-workspace.cloud.databricks.com*"},
```
arpitjasa-db, line 141: Can we remove this one as well then?

tests/test_gitlab.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.mark.parametrize("cicd_platform", ["gitlab"])
+@pytest.mark.parametrize(
+    "setup_cicd_and_project,include_feature_store,include_mlflow_recipes,include_models_in_unity_catalog",
+    [        
+        ("CICD_and_Project", "no", "no", "no"),
+        ("CICD_and_Project", "no", "no", "yes"),
+        ("CICD_and_Project", "no", "yes", "no"),
+        ("CICD_and_Project", "yes", "no", "no"),
+        ("CICD_and_Project", "yes", "no", "yes"),
+        ("CICD_Only", "no", "no", "no")
```
arpitjasa-db, line 20: What about the other `CICD_only` options?

databricks_template_schema.json
valx: The plan is to get a Field engineering account on Docker Hub for Databricks. Or use the Databricks runtime one. 

Then we'll just need to update the default docker image in this project.

tests/utils.py
valx: done

databricks_template_schema.json
arpitjasa-db: Sounds good, can you file a follow-up on your end to track this? Thanks!

tests/test_gitlab.py
valx: I copied all the options from github tests.
I think for **CICD_Only** the other options are not relevant (include_feature_store, include_mlflow_recipes, include_models_in_unity_catalog are only relevant for the **Project** part!) so we leave all set to 'no'

arpitjasa-db: Sure that's fine then, thanks!

valx: on this file I've now added the test for all clouds

arpitjasa-db (approved):

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
```diff
@@ -27,6 +27,7 @@ include:
 # Deployment Target specific values for workspace
 targets:
   dev: {{ if (eq .input_include_models_in_unity_catalog `yes`)}} # UC Catalog Name {{ end }}
+    mode: development
```
pietern, line 30: @arpitjasa-db Was this intentional?

We have an integration test that confirms successful template initialization and noticed that the job names now are prefixed with the username (DAB default for `mode: development`), so checking.

valx: I intentionally added it as I thought it would be a best practice and most environment I've been working with would benefit from it (as multiple developers could be deploying the same project into dev).
But if that has any negative impact the change could be reverted

pietern: It is, no need to revert! It seemed like a piggybacked change, so want to make sure there isn't any docs that need to be updated, or otherwise other downstream changes that need to be made.

arpitjasa-db: Yup intentional! Didn't realize this would break the naming downstream, thanks for flagging. For docs though, we don't really mention the file names for the dev target so we should be good on that

pietern: Thanks for confirming.

Issue #168, opened 2024-08-19
Bump Python version to 3.10 in workflow files


johnsonphil:
Workflow files "run-checks" and "run-tests" use Python 3.8 and "ci-tests" uses Python 3.9. Python 3.8 LTS end 10/2024.

Recommend bumping to Python 3.10 that has LTS through 10/2026

arpitjasa-db:
@johnsonphil would you mind creating a PR and confirming that these workflow files still work as expected?

mniels17:
Looks like it's only in the run-tests workflow.  We've already updated on our end for GH actions, and it works as expected.  I don't have an environment to test the ADO pipelines but I believe it also uses python 3.8 in the Azure pipelines.  I left the Azure pipelines as is.

arpitjasa-db:
@mniels17 solved this in https://github.com/databricks/mlops-stacks/pull/178, thank you!

PR #178, opened 2024-10-30
Update python version from 3.8 to 3.10 in run-tests workflow

mniels17:
Python version 3.8 is now EOL.  This resolves issue #168


mniels17:
@arpitjasa-db found some deprecated scikit-learn code and updated it in the unit_tests.

arpitjasa-db (approved):

PR #179, opened 2024-11-05
explicitly set data security mode to SINGLE_USER

shichengzhou-db:
Title.
- Set to SINGLE_USER so that mlops stack cluster creation is unaffected by default values of data_security_mode.
- Fix broken tests by upgrading github actions versions

Manual tested all workflows on all clouds (aws/azure/gcp), no failures related to data_security_mode. Although the workflows seems to be broken for various issues that need to be looked into separately.

### DIFF
--- a/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/resources/batch-inference-workflow-resource.yml.tmpl
+++ b/template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/resources/batch-inference-workflow-resource.yml.tmpl
@@ -3,6 +3,7 @@ new_cluster: &new_cluster
     num_workers: 3
     spark_version: 15.3.x-cpu-ml-scala2.12
     node_type_id: {{template `cloud_specific_node_type_id` .}}
+    data_security_mode: "SINGLE_USER"
     custom_tags:
       clusterSource: mlops-stacks_{{template `stacks_version` .}}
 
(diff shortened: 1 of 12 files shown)


template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/resources/batch-inference-workflow-resource.yml.tmpl
```diff
@@ -3,6 +3,7 @@ new_cluster: &new_cluster
     num_workers: 3
     spark_version: 15.3.x-cpu-ml-scala2.12
     node_type_id: {{template `cloud_specific_node_type_id` .}}
+    data_security_mode: "SINGLE_USER"
```
arpitjasa-db, line 6: Can we confirm these workflows still run on AWS, Azure, and GCP?

arpitjasa-db (commented):
Thanks @shichengzhou-db ! Would you mind also fixing the tests that are failing?

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/resources/batch-inference-workflow-resource.yml.tmpl
shichengzhou-db: confirmed this works on dogfood(aws), will test the other in a bit

shichengzhou-db:
> Thanks @shichengzhou-db ! Would you mind also fixing the tests that are failing?

@arpitjasa-db 
Fixed broken tests, and tested the feature engineering workflow on all clouds, the gcp one seems to have been failing before this.

(Will test the other workflows on all clouds as well tmmrw and document in the PR description)

arpitjasa-db (approved):
Thanks @shichengzhou-db !

PR #185, opened 2024-12-10
Add bundle UUID and other metadata

arpitjasa-db:
Updated off of https://github.com/databricks/mlops-stacks/pull/170


template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
```diff
@@ -1,5 +1,9 @@
 # The name of the bundle. run `databricks bundle schema` to see the full bundle settings schema.
 bundle:
+  # Do not modify the below two lines, these autogenerated fields are used by the Databricks backend.
+  uuid: {{ bundle_uuid }}
+  template: mlops-stacks
```
shreyas-goenka, line 5: No need for this, the template name will  be inferred at runtime.

shreyas-goenka (commented):
Thanks! Good to go once we remove the template arg.

template/{{.input_root_dir}}/{{template `project_name_alphanumeric_underscore` .}}/databricks.yml.tmpl
arpitjasa-db: Oh how will this be done? Are we storing some metadata after initialization to identify the template?

PR #205, opened 2025-07-16
Agentops stacks v1

alexbaur:
# AgentOps initial integration into MLOps Stacks

## Overview
This PR adds in the AgentOps stacks files to the project template and incorporates copies of existing files appended with _agentops for comparison as we continue to shape up the LLMOps flavor.

I've also added the skip commands for all files in the template folders that pertain to AgentOps automatically.

__Note:__ there are some files above the template in this PR that are _agentops versions of existing files that will need to be rectified. They are there currently for reference.


template/update_layout.tmpl
```diff
@@ -2,6 +2,75 @@
 {{ $project_name_alphanumeric_underscore := (regexp `-`).ReplaceAllString ((regexp `[^A-Za-z0-9_-]`).ReplaceAllString .input_project_name ``) `_` -}}
 {{ $root_dir := .input_root_dir}}
 
+# Skip all Agentops files in the template
+{{ skip (printf `%s/%s/%s` $root_dir $project_name_alphanumeric_underscore `data_preparation`) }}
+{{ skip (printf `%s/%s/%s` $root_dir $project_name_alphanumeric_underscore `data_preparation/data_ingestion`) }}
```
arpitjasa-db: Skipping the directory will skip all of its files, no need to repeat

arpitjasa-db (commented):
I think this is a good first step. Next step would be to add a variable in the template JSON schema, and merge the overlapping files that right now we just added a suffix too. But this can be done in a follow-up PR to update this branch

Noticed we don't have any CICD files today? Also, currently tests are failing, which is fine for now, but we will need to fix before merging to main branch

template/update_layout.tmpl
alexbaur: modified update_layout to skip the directories where needed.

